PR #83, opened 2019-04-24
Clarifying differences ui vs rest

J12934:
Added clarification for the docs on how the execution of securityTests differ when they are started via the ui vs when they are started via the api


J12934:
[User Guide 📝: Expand Persistence Docs](https://trello.com/c/1wN5PW6H/252-user-guide-%F0%9F%93%9D-expand-persistence-docs)

PR #101, opened 2020-01-20
Stabilize SSH Integration Tests

fuhrmeistery:
SSH Integrationtest were failing from time to time.
To improve the stability for these tests we introduced a new test service and updated the integrationtests


J12934:
Could you upload the `Dockerfile` used to build the `securecodebox/ssh-demo` image into the `demo` folder? Best into its own directory like the [unsafe-https](https://github.com/secureCodeBox/secureCodeBox/tree/master/demo/unsafe-https) one.

This way we can rebuild and update the image easier if necessary.

J12934:
You'll also need to add the `dummy-ssh` container to the list of services started by docker compose in travis. We don't start all by default as this would overwhelm the travis job.

PR #116, opened 2020-03-30
write adr for combined scanner

dpatanin:


dpatanin:
Please correct me if this was not the exact reasoning we had back then, when we decided on how to implement the combined processes.

Weltraumschaf:
The decission part sounds more than the context for me: It describes what drives us to the decission. And the decission itself is not very clear. To me it's not obvious what we decided here.

In the consequences part you should outline more what are the pros and the cons. Eg. one con is that we can not simply "paint" these multiple scanner pipelines in the BPMN editor.

dpatanin:
> The decission part sounds more than the context for me: It describes what drives us to the decission. And the decission itself is not very clear. To me it's not obvious what we decided here.
> 
> In the consequences part you should outline more what are the pros and the cons. Eg. one con is that we can not simply "paint" these multiple scanner pipelines in the BPMN editor.

I reworked it a little. I hope it is clearer now, but I'm afraid that i cannot dive too deep into the explicit reasoning since this decision was made about 9 months or so ago. I hope @rseedorff can add to it a little bit when reviewing.

Issue #136, opened 2020-10-07
🏗 Merge v2 Code into regular secureCodeBox Repository


J12934:
Before we can properly release the v2.0.0 we have to merge over the code of [this repository](https://github.com/secureCodeBox/secureCodeBox-v2) into the regular [secureCodeBox Repository](secureCodeBox/secureCodeBox-v2).

Things that should be thought of (list probably incomplete):

- [x] Update go namespace (#144)
- [x] Search for references in the docs and code (#144, partly docs missing)
- [x] Ensure that the Docker Builds still works (Including github action secrets)
- [x] Ensure that the Helm Publishing still works (Including github action secrets)
- [x] Copy Netlify webhook (configured, works)
- [x] Update Codacity integration
- [x] Update CodeClimate integration
- [x] Update fossa integration

J12934:
@rseedorff can you take a look at the fossa / codacity settings?

J12934:
Should be all done 🎉

PR #46, opened 2018-09-19
Added documentation for starting a zap scan without spider

MartinLang1:
Documentation for functionality introduced in corresponding pull request of the "engine" repository

PR #144, opened 2020-10-14
Update References to Github Repository

J12934:
Resolves the following points from #136:

- [ ] Update go namespace
- [ ] Search for references in the docs and code

### DIFF
--- a/hooks/declarative-subsequent-scans/templates/NOTES.txt
+++ b/hooks/declarative-subsequent-scans/templates/NOTES.txt
@@ -10,4 +10,4 @@ $ kubectl get cascadingrules
 You need to explicitly turn on scan cascading for every scan you use.
 You can do that by setting a label selector which matches all rules you want to use.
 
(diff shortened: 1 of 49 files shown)


hooks/declarative-subsequent-scans/templates/NOTES.txt
```diff
@@ -10,4 +10,4 @@ $ kubectl get cascadingrules
 You need to explicitly turn on scan cascading for every scan you use.
 You can do that by setting a label selector which matches all rules you want to use.
 
-Find out more, on the docs: TODO(https://github.com/secureCodeBox/secureCodeBox-v2/issues/46)
```
fuhrmeistery, line 13: We should not update this link because this would break the link (rerouting it to https://github.com/secureCodeBox/secureCodeBox/pull/46)

Weltraumschaf (approved):

hooks/declarative-subsequent-scans/templates/NOTES.txt
J12934: Oh good catch 👍
Will fix post merge

PR #165, opened 2020-10-19
Improved (Un-)Install scripts

SebieF:
1. Improved the uninstall-script.sh to run through all commands even if some programs have already been uninstalled
2. Created an interactive-install.sh to help mostly inexperienced users with their first install. The script goes through all available scanners, demo-apps and hooks available in our respective directories and asks the user if he wants to install the available scanner/app/hook. Also asks for a custom namespace for the demo-apps.


bin/interactive-install.sh
```diff
@@ -0,0 +1,140 @@
...
+else
+  echo -e "\e[31mKubectl not found, quitting..\e[0m"
+  exit
+fi
+
+echo
+echo -e "Creating namespace securecodebox-system"
+kubectl create namespace securecodebox-system || echo -e "Namespace already exists.. "
+
+if [[ $PWD == *"bin" ]]; then
```
Weltraumschaf, line 31: Here you should use a more reliable way. It is possible that someone executes this script from somewhere else, eg.:

```bash
cd /opt/local/bin
/path/to/repo/bin/interactive-install.sh
```

A good reliable way to find out where you are:

```bash
# @see: http://wiki.bash-hackers.org/syntax/shellvars
[ -z "${SCRIPT_DIRECTORY:-}" ] \
    && SCRIPT_DIRECTORY="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )" \
    && export SCRIPT_DIRECTORY
```

This also works if someone softlinks your script (eg. into `$HOME/bin`). You can use `dirname($SCRIPT_DIRECTORY)` as base direcotry.

bin/interactive-install.sh
```diff
@@ -0,0 +1,140 @@
...
+fi
+echo -e "Installing the operator in the securecodebox-system namespace"
+helm -n securecodebox-system upgrade --install securecodebox-operator ./operator/ || echo -e "\e[31mOperator installation failed, cancelling..e[0m"
+echo -e "\e[32mSuccessfully installed the operator!\e[0m"
+
+echo
+echo -e "Starting to install scanners.."
+cd scanners
+scanners=()
+for directory in */ ; do
```
Weltraumschaf, line 42: Here you should use an absolute path derived from the above mentioned base directory. Why? Here you completely rely that you are in the right directory, but due to edge cases you may end up somewhere else and slurp in everything listed in that location. So this may evolve to kinda code execution, when I achieve to trick this script iterating over a directory where I put my malicious helm charts.

bin/interactive-install.sh
```diff
@@ -0,0 +1,140 @@
...
+do
+  while true
+  echo -e "Do you want to install $scanner? [y/n/(r)eadme]"
+  do
+    read -r line
+    if [[ $line == *[Yy] ]]; then
+      scannerName="${scanner//+([_])/-}" # Necessary because ssh_scan is called ssh-scan
+      helm upgrade --install "$scannerName" ./"$scanner"/ || echo -e "\e[31mInstallation of ""$scannerName"" failed\e[0m"
+      break
+    elif [[ $line == *[r] ]]; then
```
Weltraumschaf, line 56: For what is this branch for?

bin/interactive-install.sh
```diff
@@ -0,0 +1,140 @@
...
+namespace="default"
+if [[ $line == *[Yy] ]]; then
+  echo -e "Please provide a name for the namespace:"
+  read -r namespace
+  kubectl create namespace "$namespace" || echo -e "Namespace already exists.. "
+fi
+
+cd ../demo-apps
+demos=()
+for directory in */ ; do
```
Weltraumschaf, line 80: Same problem like mentioned above with the scanners.

bin/interactive-install.sh
```diff
@@ -0,0 +1,140 @@
...
+for demo in "${demos[@]}"
+do
+  while true
+  echo -e "Do you want to install $demo? [y/n/(r)eadme]"
+  do
+    read -r line
+    if [[ $line == *[Yy] ]]; then
+      helm upgrade --install -n "$namespace" "$demo" ./"$demo"/ || echo -e "\e[31mInstallation of ""$demo"" failed\e[0m"
+      break
+    elif [[ $line == *[r] ]]; then
```
Weltraumschaf, line 93: For what is this branch for?

bin/interactive-install.sh
```diff
@@ -0,0 +1,140 @@
...
+done
+
+echo
+echo -e "\e[32mCompleted to install demo-apps!\e[0m"
+
+echo
+echo -e "Starting to install hooks.."
+cd ../hooks
+hooks=()
+for directory in */ ; do
```
Weltraumschaf, line 108: Same as above.

bin/interactive-install.sh
```diff
@@ -0,0 +1,140 @@
...
+for hook in "${hooks[@]}"
+do
+  while true
+  echo -e "Do you want to install $hook? [y/n/(r)eadme]"
+  do
+    read -r line
+    if [[ $line == *[Yy] ]]; then
+      helm upgrade --install "$hook" ./"$hook"/ || echo -e "\e[31mInstallation of ""$hook"" failed\e[0m"
+      break
+    elif [[ $line == *[r] ]]; then
```
Weltraumschaf, line 120: Same as above

Weltraumschaf (changes requested):
The parts for installing scanners, hooks and demo-apps is quite repetitive. You should consider using bash function to reduce duplicate code.

bin/interactive-install.sh
```diff
@@ -0,0 +1,109 @@
...
+  for resource in "${resources[@]}"
+  do
+  while true
+  echo -e "Do you want to install $resource? [y/n/(r)eadme]"
+  do
+    read -r line
+    # Install:
+    if [[ $line == *[Yy] ]]; then
+      resourceName="${resource//+([_])/-}" # Necessary because ssh_scan is called ssh-scan
+      helm upgrade --install -n "$2" "$resourceName" ./"$resource"/ || echo -e "\e[31mInstallation of ""$resource"" failed\e[0m"
```
rseedorff, line 60: It would be great to add an configuration option to define a namespace where to install the resource as already implemented for the demo-apps

bin/interactive-install.sh
```diff
@@ -0,0 +1,109 @@
...
+
+[ -z "${SCRIPT_DIRECTORY:-}" ] \
+  && SCRIPT_DIRECTORY="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )" \
+  && export SCRIPT_DIRECTORY
+
+cd "$SCRIPT_DIRECTORY"
+cd ..
+
+echo -e "Installing the operator in the securecodebox-system namespace"
+helm -n securecodebox-system upgrade --install securecodebox-operator ./operator/ && echo -e "\e[32mSuccessfully installed the operator!\e[0m" \
```
rseedorff, line 39: It would be great to add an configuration option to decide if all following ressources are installed via a local cloned repo or instead via the public helmChart lib like documented in our documentation.

bin/uninstall-all.sh
```diff
@@ -2,28 +2,28 @@
...
-helm uninstall aah
-helm uninstall gwh
+helm uninstall aah || true
+helm uninstall gwh || true
 
-helm uninstall elkh
+helm uninstall elkh || true
 
-kubectl delete namespaces securecodebox-system
+kubectl delete namespaces securecodebox-system || true
```
rseedorff, line 29: If we provide an uninstall script it should also contain the uninstallation of all CRDs like documented here: https://docs.securecodebox.io/docs/getting-started/uninstallation#delete-roles-rolebindings-and-serviceaccounts

bin/interactive-install.sh
```diff
@@ -0,0 +1,109 @@
+#!/usr/bin/env bash
+
+set -eu
+shopt -s extglob
+
+echo -e "\e[32mWelcome to the secureCodeBox!\e[0m"
+echo -e "This interactive installation script will guide you through all the relevant installation steps in order to have you ready to scan."
+
+echo -e "Start ? [y/n]"
```
rseedorff, line 9: It would be great if you add some additional CLI flags in the case i want to install everything without the interactive dialog, eg `--all` with our defaults (for namespaces). This would help for example to use this script in an vagrant context where we could provide a fully all-in-one vagrant box with autoinstalling everything (minikube + SCBv2)

bin/interactive-install.sh
```diff
@@ -28,41 +28,55 @@ echo
...
 
-if [[ $PWD == *"bin" ]]; then
-  cd ..
-fi
+[ -z "${SCRIPT_DIRECTORY:-}" ] \
+  && SCRIPT_DIRECTORY="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )" \
+  && export SCRIPT_DIRECTORY
+
+cd "$SCRIPT_DIRECTORY"
+cd ..
```
Weltraumschaf, line 36: That's not very intuitive for an experiences shell script reade ;-)

A good patter is to define a base directory and work with that:

```bash
BASE_DIR=$(dirname "${SCRIPT_DIRECTORY}")
```

From now on always use `"${BASE_DIR}/path/to/my/stuff"` style paths. And always escape paths with quotes and expansion to prevent unwanted globbering or wordsplitting: `"${VAR}"`.

Weltraumschaf (commented):
I suggest a pair coding session with me.

bin/interactive-install.sh
SebieF: Should the question for another namespace be displayed for every single resource (scanner, hook, demo-app) or for all scanners, hooks and demo-apps combined (like it is at the moment for the demo-apps)?

bin/install.sh
```diff
@@ -0,0 +1,136 @@
+#!/usr/bin/env bash
+
+# Official installation script for the secureCodeBox
```
Weltraumschaf, line 3:
```suggestion
# Official installation script for the secureCodeBox
#
```

bin/install.sh
```diff
@@ -0,0 +1,136 @@
...
+# Creates namespace securecodebox-system and installs the operator
+# Then interactively installs all possible resources (scanners, demo-apps, hooks)
+# https://docs.securecodebox.io/
```
Weltraumschaf, line 6:
```suggestion
# Creates namespace, securecodebox-system, and installs the operator.
# Then interactively installs all possible resources (scanners, demo-apps, hooks).
#
# For more information see https://docs.securecodebox.io/
```

bin/install.sh
```diff
@@ -0,0 +1,136 @@
...
+
+COLOR_HIGHLIGHT="\e[35m"
+COLOR_OK="\e[32m"
+COLOR_ERROR="\e[31m"
+COLOR_RESET="\e[0m"
+
+# @see: http://wiki.bash-hackers.org/syntax/shellvars
+[ -z "${SCRIPT_DIRECTORY:-}" ] \
+  && SCRIPT_DIRECTORY="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )" \
+  && export SCRIPT_DIRECTORY
```
Weltraumschaf, line 19:
```suggestion
  && SCRIPT_DIRECTORY="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
```
Learned this weekend the hard way that exporting this is usually a bad idea: It overrides the same variable which may be set in an included script and then this subsequent script may fail.

bin/uninstall.sh
```diff
@@ -0,0 +1,40 @@
...
+# Removes all available resources (scanners, demo-apps, hooks, operator) and namespaces
+# https://docs.securecodebox.io/
+
+set -eu
+shopt -s extglob
+
+# @see: http://wiki.bash-hackers.org/syntax/shellvars
+[ -z "${SCRIPT_DIRECTORY:-}" ] \
+  && SCRIPT_DIRECTORY="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )" \
+  && export SCRIPT_DIRECTORY
```
Weltraumschaf, line 13:
```suggestion
  && SCRIPT_DIRECTORY="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
```

see above

Weltraumschaf (changes requested):

Weltraumschaf (approved):

Issue #171, opened 2020-10-21
⚓️ New persistenceProvider Hook for OWASP DefectDojo


rseedorff:
## New Hook implementation request

**Is your feature request related to a problem? Please describe.**
As a User i would like to use the [OWASP DefectDojo Projekt](https://www.owasp.org/index.php/OWASP_DefectDojo_Project) to analyse my findings (_instead or in parallel to my kibana dashboard_). While the secureCodeBox has its major focus on the automation part of the security scanner execution, defectDojo is good in visualising, deduplication and analysing the findings.

To integrate OWASP DefectDojo it's necessary to implement a new Hook which pushes the raw finding results of each scanner to the [DefectDojo API: ImportScan](https://app.swaggerhub.com/apis/DefectDojo/defect-dojo_api_v_2/1.0.0#/import-scan/import-scan_create). DefectDojo can be started as a separate Docker Container or via HelmChart in a K8S cluster environment.

Possible Scanner Integrations:
- NMAP (XML output (use -oX))
- Nikto (XML output)
- ZAP (ZAP XML report format)

**Describe the solution you'd like**
The is already a working integration solution implemented for the secureCodeBox V1, which maybe can be migrated or reused in the secureCodeBox V2. Therefore the existing code must be wrapped in a Hook.

**Describe alternatives you've considered**
Another alternative could be implement this hook with a complete green field approach. But this alternative would take much longer and has no real benefits.

**Additional context**
- https://github.com/secureCodeBox/engine/tree/master/scb-persistenceproviders/defectdojo-persistenceprovider
- https://github.com/secureCodeBox/engine/tree/feature/defect-dojo-config-extension/scb-persistenceproviders/defectdojo-persistenceprovider

## Steps to implement a new Hook

- [x] Create a new folder with the name of the [hook here](https://github.com/secureCodeBox/secureCodeBox/tree/master/hooks)
- [x] Add a README and give a brief overview of the scanner and its configuration options.
- [x] Add (optional) a Dockerfile for the scanner if there is no existing one publicly available on dockerHub
- [x] Use the [Hook-SDK](https://github.com/secureCodeBox/secureCodeBox/tree/master/hook-sdk) to implement a new hook (currently based on NodeJS)
- [x] Add unit tests with at minimum 80% test coverage
- [ ] Add some example scan.yaml and finding.yaml files in the example folder
[...]

PR #176, opened 2020-10-22
Switch ZAP Output Format to XML

J12934:
This PR switches the output format used by the ZAP ScanTypes to use XML.

This is done to enable the coming DefectDojo Persistent Provider (see #171) to use the native DefectDojo ZAP import rather than falling back to import them as generic findings.

The secureCodeBox findings outputted should remain the same, regardless of the change.


fuhrmeistery (approved):

fuhrmeistery:
How do we want to handle the codeclimate issue?

J12934:
@fuhrmeistery I don't really know how to improve the parse function anymore. I'm guessing that its getting this bad of a score because of the three nested loops. These, at least in my opinion, these loops belong together and wouldn't really make sense to split these up in separate functions.

PR #182, opened 2020-10-26
Adding unattended mode to install script

SebieF:
Easier use in CLI for the install.sh to install every available resources via --all
1. Extracting a lot of functions to avoid duplicated code
2. Read parameters from command line
3. Give usage feedback if not properly used
4. Enhanced docu and fixing some minor code issues like inconsistent local variables

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -8,6 +8,8 @@
 # There exist two modes:
 # Call without parameters to install interactively
 # Call with --all to install all available resources automatically
+# Call with --scanners / --demo-apps / --hooks to only install the wanted resources
+# Call with --help for usage information
 #
 # For more information see https://docs.securecodebox.io/
 
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -3,7 +3,11 @@
 # Official installation script for the secureCodeBox
 #
 # Creates namespace, securecodebox-system, and installs the operator.
-# Then interactively installs all possible resources (scanners, demo-apps, hooks).
+# Then installs all possible resources (scanners, demo-apps, hooks).
+#
+# There exist two modes:
+# Call without parameters to install interactively
+# Call with --all to install all available resources automatically
 #
 # For more information see https://docs.securecodebox.io/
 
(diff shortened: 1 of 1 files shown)


bin/install.sh
```diff
@@ -22,21 +26,57 @@ COLOR_RESET="\e[0m"
...
     echo "${message}"
-  elif [[ $# = 2 ]]; then
+  elif [[ $# == 2 ]]; then
     local color="${1}"
     local message="${2}"
     echo -e "${color}${message}${COLOR_RESET}"
   fi
 }
 
+function usage() {
```
Weltraumschaf, line 41: Usually the usage string looks like: `Usage: command --required-option [--optional-options] [-h|--help] <filename>` so in this case I would expcet: `Usage: $(basename "$0") [--all] [--scanners] [--hooks] [--demo-apps] [-h]`.

A good pattern is to store usage and help in variables, eg.:

```
USAGE="Usage: $(basename "$0") <hostname> <jenkins_secret>"
HELP=$(cat <<- EOT
Here tell the user that w/o arguments the installatiuon is interactive.

Options:
  --all                  Explanation...
  --scanners       Explanation
  --hooks            Explanation
  --demo-apps   Explanation
  -h|--help          Explanation

Examples:

Add some examples.
EOT
)
```

These variables you can use in a function to print help:

```
print_help() {
    echo "$USAGE"
    echo
    echo "$HELP"
    echo
}
```

The `USAGE` you can print before any non-zero exit as hint.

bin/install.sh
```diff
@@ -46,92 +86,111 @@ function installResources() {
...
+print "$COLOR_HIGHLIGHT" "  ___  ___  ___ _   _ _ __ ___| |     ___   __| | ___| |_) | _____  __       "
+print "$COLOR_HIGHLIGHT" " / __|/ _ \/ __| | | | '__/ _ \ |    / _ \ / _  |/ _ \  _ < / _ \ \/ /       "
+print "$COLOR_HIGHLIGHT" " \__ \  __/ (__| |_| | | |  __/ |___| (_) | (_| |  __/ |_) | (_) >  <        "
+print "$COLOR_HIGHLIGHT" " |___/\___|\___|\__,_|_|  \___|\_____\___/ \__,_|\___|____/ \___/_/\_\       "
+print "$COLOR_HIGHLIGHT" "                                                                             "
+
+if [[ $# == 0 ]]; then
+    interactiveInstall
+  elif [[ $# == 1 ]]; then
+    if [[ $1 == "--all" ]]; then
```
Weltraumschaf, line 191: What about the other options (--scanners etc.)?

Weltraumschaf (changes requested):

Weltraumschaf (approved):
LGTM

PR #191, opened 2020-11-03
Vagrant install

Weltraumschaf:
Now we have the unattended install script which is now added to the Vagrant setup.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -7,8 +7,6 @@ Vagrant.configure("2") do |config|
 
   config.vm.box = "debian/buster64"
   config.vm.hostname = "securecodebox"
-  # Don't sync anything onto the box.
-  config.vm.synced_folder base_dir, '/vagrant', disabled: true
 
   # We use the same defaults like Docker Desktop.
   memory = 2048
@@ -43,6 +41,8 @@ Vagrant.configure("2") do |config|
   # Redirecting STDERR to /dev/null because Minikube print download progress
   # for the images to STDERR which clutters up the Vagrant output w/ error output!
   config.vm.provision :shell, privileged: false, inline: "sudo -g docker minikube start 2>/dev/null"
+  # Install everything from secureCodeBox:
+  config.vm.provision :shell, privileged: false, inline: "sudo -g docker /vagrant/bin/install.sh --all"
 
   # Do not automatically install VirtualBox guest additions, if available.
   # Because this would take lot of time with additional reboot.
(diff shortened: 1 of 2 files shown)


bin/install-minikube.sh
```diff
@@ -1,32 +1,83 @@
 #!/usr/bin/env bash
 
+#
+# Install the Minikube setup in the all-in-one Vagrant box.
+#
+# This scriptmust be executedas root.
```
SebieF, line 6: Typo: executed as

bin/install-minikube.sh
```diff
@@ -1,32 +1,83 @@
...
+update_system
+add_docker_apt_source
+add_kubectl_apt_source
+add_helm_apt_source
+
+apt-get -y update
+apt-get install -y \
+  docker-ce \
+  kubectl \
+  helm
```
SebieF, line 78: Maybe extract to separate function "install"

Vagrantfile
```diff
@@ -43,6 +41,8 @@ Vagrant.configure("2") do |config|
   # Redirecting STDERR to /dev/null because Minikube print download progress
   # for the images to STDERR which clutters up the Vagrant output w/ error output!
   config.vm.provision :shell, privileged: false, inline: "sudo -g docker minikube start 2>/dev/null"
+  # Install everything from secureCodeBox:
+  config.vm.provision :shell, privileged: false, inline: "sudo -g docker /vagrant/bin/install.sh --all"
```
SebieF, line 45: sudo could be avoided if docker groups are used, but I'm not sure if that makes sense in this context

SebieF (approved):

bin/install-minikube.sh
Weltraumschaf: IMHO too much overhead because there is no redundancy and a extra function is no good here.

Vagrantfile
Weltraumschaf: No couldn't be avoided. All provisioners run in the same login session. So at this point the vagrant user was added to the group docker (see install-minikube.sh#83), but this has no effect until the user relogin. This is not possible here. So the workarround is (as documented in line 40) to use `sudo -g` to execute the command with the docker group. So this does exactly what you recommend.

SebieF (approved):

Weltraumschaf:
Merged this as admin because reviewer approved it. GH has some problems w/ my signature :/

PR #195, opened 2020-11-05
Fix Helm2 Chart Containing the Packaged Helm3 Chart

J12934:
This fixes a problem where the packaged helm3 chart .tgz is included in the packaged helm2 chart .tgz.

This could potentially lead to sizing problems.

This pr also exclude various other "code" files from the operator chart, which aren't really required there.


rseedorff (approved):
Good catch!

Issue #193, opened 2020-11-04
Add a CLA or DCO to the project


rseedorff:
Add a Contributor License Agreement (CLA) or Developer Certificate of Origin (DCO) to the project

Background Informations:
- https://medium.com/@flamefew/clas-and-using-dco-clearly-e46b09a4c048
- https://opensource.com/article/18/3/cla-vs-dco-whats-difference

Todos:

- [ ]  ~~Ask apache foundation~~
- [ ]  ~~Ask a lawyer~~
- [x]  Ask https://fsfe.org/

Weltraumschaf:
This relates to #194: If we choose DCO we should use license headers in every file. That's stated by one of linked article.

fuhrmeistery:
To enable us to use DCO we will need to:
- Add License Headers to each file
- Require every commit in every new PR to include "Signed-off-by" (could be automated by [Probot](https://probot.github.io/apps/dco/))

Stated by the article above we may additionally need:
- A `CONTRIBUTORS` file
- A DCO Copy with SPDX Licenses that are considered appropriate

Weltraumschaf:
Mailed the FSFE. Today they requested the mail translated in English :/

Weltraumschaf:
I got the first answer from FSFE Legal Coordinator:

>> 1. Why is it necessary to add a license header to each file? We know that lot of OSS project do this, but we didn't found a solid legal reason which explains why just one single "license.txt" is not sufficient.
>
> This is not strictly a legal requirement, however it is a practical one,
> and the FSFE has found to be good practice for downstream projects. In
> many software projects there might be more than one license at play; not
> only the license you choose to cover your project, but also the licenses
> that apply to code or components that are reused into your project. In
> many situations, these may not be the same licenses.
> 
> For example, let's say that a project (Project A) chooses to use the MIT
> license overall. However, it reuses components that are covered by
> another license, for example, the Apache license. These components in
> Project A's repository will therefore have different copyright holders,
> and a different applicable license, than the rest of the files. It is
> therefore very useful to mark out for each file who the copyright owner
> and which license applies. If a developer wants to reuse components or
> source code from Project A later down the line, they may not be able to
> immediately tell that the Apache components have a different copyright
> holder and are licensed under the Apache license. They may instead
> assume that it is licensed under the MIT license, and that the copyright
> belongs to the Project A's developers. Having the copyright and license
> information readily available in each file reduces this kind of
[...]

PR #199, opened 2020-11-11
Finding Post Processing

### LINKED ISSUE #131: ⚓️ Add NMAP Findings post processing to update the severity of port findings
As a secureCodeBox user i would like to be able to change / manipulate the nmap findings severity, based on a simple rule pattern. By default all NMAP Port findings are classified with the severity *Informational*. In some use cases i would like to change the port specific severity classification based on custom rules. For example if some port ranges are prohibited 🚫 to use.

The custom severity configuration should also be applied if the NMAP scan will be started by *cascadingScanRules* (e.g. as subsequent scan started by AMASS).

In the secureCodeBox v1 there is an NMAP specific implementation example how to change some port specific severity classifications, but without any configuration options: https://github.com/secureCodeBox/engine/commit/c40b499c9947b5d3b601c5ba43427a0d781c01b4

An example configuration could look something like this:
```yaml
openport:
  port: 21
    severity: HIGH
  port: 22
    severity: MEDIUM
  port: 389
    severity: HIGH
  port: 9200
    severity: LOW
```

The *updatefield-hook* implements a lightweight findings post processing. But currently it is only possible to add new attributes to findings. But it has no feature to manipulate finding attributes based on rules. Extending the updatefield-hook in this way would maybe leed to a more generic solution which could also apply to other scanners and usecases.

For a more generic approach maybe the cascadingRule definition configuration semantic could be a good example how to configure the post processing:

```yaml

rules:
  - rule:
       matches:
         anyOf:
           - category: "Open Port"
              attributes:
                port: 21
                state: open
           - category: "Open Port"
              attributes:
                port: 389
                state: open
        attribute:
          name: "severity"
          value: "high"
  - rule:
       matches:
         anyOf:
           - category: "Open Port"
              attributes:
                port: 22
                state: open
        attribute:
          name: "severity"
          value: "medium"
  - rule:
       matches:
         anyOf:
           - category: "Open Port"
              attributes:
                port: 9200
                state: open
        attribute:
          name: "severity"
          value: "low"
```

fuhrmeistery:
If applied this PR will
- Add a hook to post process findings
- Add tests for post process hook
- Add docs for post process hook
- Add Basic Integration Test for Hook
- Include Image build for post process hook in CI
- close #131


hooks/scan-postprocessing/hook.js
```diff
@@ -0,0 +1,56 @@
...
+module.exports.handle = handle;
+/**
+ * Goes thought the Findings and the Scan Postprocessing Rules
+ * and applies the changes to the findings defined in the rules if matching
+ */
+function applyRules(rules, findings) {
+  const newFindings = []
+  findings.forEach(finding => {
+    let processedFinding = {};
+    Object.assign(processedFinding, finding);
```
J12934, line 35: What is this for? Is it really required to copy the finding?

hooks/scan-postprocessing/hook.js
```diff
@@ -0,0 +1,56 @@
...
+    rules.forEach(rule => {
+      const isRuleMatching = rule.matches.anyOf.some(condition => isMatch(processedFinding, condition));
+      if (isRuleMatching) {
+        processedFinding = postprocessingFindings(processedFinding, rule);
+      }
+    })
+    if (notEqual(processedFinding, finding)) {
+      newFindings.push(processedFinding);
+    }
```
J12934, line 44: This would drop all findings **not** matched by a rule, right?
Is this intended? I'd have assumed that non matched findings would stay unchanged.

hooks/scan-postprocessing/hook.js
```diff
@@ -0,0 +1,56 @@
...
+  });
+  return newFindings
+}
+
+function postprocessingFindings(finding, rule) {
```
J12934, line 49: Would prefer to name this `postProcessFinding`

hooks/scan-postprocessing/hook.js
```diff
@@ -0,0 +1,56 @@
...
+  const newFinding = Object.assign(finding, rule.override);
```
J12934, line 50: Might be better to use https://lodash.com/docs/4.17.15#merge for this as Object.assign doesn't seem to "merge" recursively which might be a problem for the `attributes` field.

Would be cool to add a test which uses a rule with overrides for field in the `attributes` object.

hooks/scan-postprocessing/hook.js
J12934: Also would be a good candidate for a new test which includes findings **not** matched by a rule.

J12934 (changes requested):
Great work so far 👏

Would it maybe be better to call this finding post processing instead of scan post processing, as this really only changes findings, not scans?

hooks/scan-postprocessing/hook.js
```diff
@@ -0,0 +1,56 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+ */
+const { isMatch, isEqual } = require("lodash")
+async function handle({
+  getFindings,
+  updateFindings,
+  rules = process.env["RULES"],
```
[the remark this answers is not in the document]
fuhrmeistery: This is currently not possible due to the CRD of Hooks

hooks/finding-post-processing/README.md
```diff
@@ -0,0 +1,40 @@
...
+---
+
+<!-- end -->
+
+## Deployment
+
+Installing the _Scan Postprocessing hook will add a ReadAndWrite Hook to your namespace, which can be used to add or update fields from your findings meeting specified conditions.
+
+```bash
+helm upgrade --install spp secureCodeBox/scan-postprocessing
```
SebieF, line 16: Shouldn't this be 
```bash
helm upgrade --install fpp ./hooks/finding-post-processing
```
?

hooks/finding-post-processing/templates/NOTES.txt
```diff
@@ -0,0 +1,16 @@
+Copyright 2020 iteratec GmbH
```
SebieF, line 1: The header will also show on every installation with helm, is this desired/needed?

hooks/finding-post-processing/hook.js
```diff
@@ -0,0 +1,50 @@
...
+}) {
+  const findings = await getFindings();
+  const res = applyRules(rules, findings);
+  if (res.hasChanged) {
+    await updateFindings(res.findings);
+  }
+}
+module.exports.handle = handle;
+/**
+ * Goes thought the Findings and the Finding Post Processing Rules
```
SebieF, line 30: Typo: through

SebieF (commented):
Great work!

hooks/finding-post-processing/README.md
```diff
@@ -0,0 +1,40 @@
...
+type: "dataProcessing"
+state: "released"
+usecase: "Updates fields for findings meeting specified conditions."
+---
+
+<!-- end -->
+
+## Deployment
+
+Installing the _Scan Postprocessing hook will add a ReadAndWrite Hook to your namespace, which can be used to add or update fields from your findings meeting specified conditions.
```
J12934, line 13: The _Scan looks like broken markdown?

hooks/finding-post-processing/templates/NOTES.txt
fuhrmeistery: I'm not sure but I think if we want to  use DCO we should include the license header (See #193 )

tests/integration/hooks/finding-post-processing.test.js
```diff
@@ -12,6 +12,7 @@ test(
 
     expect(count).toBe(2);
     expect(severities.high === 1)
```
J12934, line 14: Does this test really fail if the condition evaluates to false? I've only ever seen expect be used together with a matcher.

I've only ever seen it used together with a matcher and the docs don't seem to say what happens when you call it by itself: https://jestjs.io/docs/en/expect#expectvalue

fuhrmeistery: You are right this was simply a mistake :+1:

hooks/finding-post-processing/templates/NOTES.txt
J12934: Can we at least comment these out?
Should be possible here by using go templates: https://godoc.org/text/template#hdr-Actions

tests/integration/hooks/finding-post-processing.test.js
```diff
@@ -0,0 +1,17 @@
+const { scan } = require('../helpers')
```
J12934, line 1: This test isn't integrated into the CI right?
Is this intentional?

tests/integration/hooks/__testFiles__/values.yaml
```diff
@@ -0,0 +1,34 @@
+# Copyright 2020 iteratec GmbH
```
J12934, line 1: File name should be specific to finding post process, as this is in a generic folder

J12934 (approved):

PR #201, opened 2020-11-13
Nikto tests

SebieF:
When applied this pull request will add a simple integration test for the nikto scanner tested against the bodgeit demo-app.

Branch is forked from sslyze-tests, must probably be considered when merging.


SebieF:
Deprecated

PR #214, opened 2020-11-18
Splitted screenshooter cascading rule into two different rules

rseedorff:
## Checklist

## Description

Splitted screenshooter cascading rule into two different rule because NMAP sometimes identifies the HTTP services on port 443 instead of HTTPS and this leads to errors.


J12934:
Wouln't this lead to two duplicated scans then?

One unsuccessful scan for http://example.com:443
And one successful one for https://example.com:443

rseedorff:
Good point 🤔

PR #215, opened 2020-11-18
Install script improvement

SebieF:
When applied this pull request will add some improvements to the existing install script:

1. Better documentation
2. Removing duplicated code, clarifying the main function
3. Separating help from usage
4. Refactoring some function names to clarify their side effects
5. Adding a check to avoid the attempt of installing non-existing resources

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -13,7 +13,7 @@
 #
 # For more information see https://docs.securecodebox.io/
 
-set -e
+set -euo pipefail
 shopt -s extglob
 
 USAGE="Usage: $(basename "$0") [--all] [--scanners] [--hooks] [--demo-apps] [--help|-h]"
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -27,6 +27,11 @@ COLOR_RESET="\e[0m"
 
 BASE_DIR=$(dirname "${SCRIPT_DIRECTORY}")
 
+INSTALL_INTERACTIVE=''
+INSTALL_SCANNERS=''
+INSTALL_DEMO_APPS=''
+INSTALL_HOOKS=''
+
 function print() {
   if [[ $# == 0 ]]; then
     echo
@@ -154,7 +159,7 @@ function installResources() {
   print "$COLOR_OK" "Completed to install '$resource_directory'!"
 }
 
-function interactiveInstall() {
+function welcomeToInteractiveInstall() {
   print "$COLOR_HIGHLIGHT" "Welcome to the secureCodeBox!"
   print "This interactive installation script will guide you through all the relevant installation steps in order to have you ready to scan."
   print "Start? [y/N]"
@@ -166,11 +171,9 @@ function interactiveInstall() {
   else
     exit
   fi
+}
 
-  exitIfKubectlIsNotInstalled
-  exitIfHelmIsNotInstalled
-  createNamespaceAndInstallOperator
-
+function interactiveInstall() {
   print
   print "Starting to install scanners..."
   installResources "$BASE_DIR/scanners" "default" False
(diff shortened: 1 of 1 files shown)


bin/install.sh
```diff
@@ -13,7 +13,7 @@
 #
 # For more information see https://docs.securecodebox.io/
 
-set -eu
+set -euo pipefail
```
Weltraumschaf, line 16: Has nothing todo with this commit.

bin/install.sh
```diff
@@ -40,27 +40,29 @@ function print() {
   fi
 }
 
+# TODO: Separate usage from help.
```
Weltraumschaf: Remove this.

bin/install.sh
```diff
@@ -40,27 +40,29 @@ function print() {
...
 function usage() {
   local usage
-  usage="Usage: $(basename "$0") [--all] [--scanners] [--hooks] [--demo-apps] [--help]"
+  usage="Usage: $(basename "$0") [--all] [--scanners] [--hooks] [--demo-apps] [--help|-h]"
```
Weltraumschaf: Has nothing todo with this commit.

bin/install.sh
```diff
@@ -40,27 +40,29 @@ function print() {
...
   local help
   help=$(cat <<- EOT
-        The installation is interactive if no arguments are provided.
```
Weltraumschaf, line 48: Has nothing todo with this commit.

bin/install.sh
```diff
@@ -278,8 +281,22 @@ print "$COLOR_HIGHLIGHT" " \__ \  __/ (__| |_| | | |  __/ |___| (_) | (_| |  __/
 print "$COLOR_HIGHLIGHT" " |___/\___|\___|\__,_|_|  \___|\_____\___/ \__,_|\___|____/ \___/_/\_\       "
 print "$COLOR_HIGHLIGHT" "                                                                             "
 
+
 if [[ $# == 0 ]]; then
     interactiveInstall
   else
     parseArguments "$@"
 fi
+
```
Weltraumschaf, line 288: Remove this.

bin/install.sh
```diff
@@ -27,6 +27,11 @@ COLOR_RESET="\e[0m"
 
 BASE_DIR=$(dirname "${SCRIPT_DIRECTORY}")
 
+INSTALL_INTERACTIVE=false
```
Weltraumschaf, line 32: In bash thereis no boolean. So this is implicit a sting. To clarify this behaviour it is good practice in bash to use `""` empty string for`false` and `"true"` as non-empty string for `true`.

bin/install.sh
```diff
@@ -206,25 +211,19 @@ function interactiveInstall() {
...
-  exitIfKubectlIsNotInstalled
-  exitIfHelmIsNotInstalled
-  createNamespaceAndInstallOperator
 
-  local install_scanners=$1
-  local install_demo_apps=$2
-  local install_hooks=$3
 
-  if [[ $install_scanners == true ]]; then
+  if [[ $INSTALL_SCANNERS == true ]]; then
```
Weltraumschaf, line 213: Here check for not-empty (s. above): `if [[ ! z- "$VAR" ]]; then`

bin/install.sh
```diff
@@ -130,7 +130,10 @@ function installResources() {
     [ -d "${path}" ] || continue # skip if not a directory
     local directory
     directory="$(basename "${path}")"
-    resources+=("${directory}")
+    # Check if directory contains Chart.yaml File:
+    if [[ $(find "$resource_directory"/"$directory"/Chart.yaml 2>/dev/null | wc -l) -gt 0 ]]; then
```
Weltraumschaf, line 134: Why is `if [[ -f "$resource_directory/$directory/Chart.yaml" ]]; then` not sufficient here?

Weltraumschaf (changes requested):

Weltraumschaf (approved):

PR #220, opened 2020-11-19
Add Configurable ttlSecondsAfterFinished Fields to Parse and Hook Jobs

### LINKED ISSUE #219: ➹ Implement ttlSecondsAfterFinished alpha feature support for all CRDs
**Is your feature request related to a problem? Please describe.**

> The TTL controller provides a TTL (time to live) mechanism to limit the lifetime of resource objects that have finished execution. TTL controller only handles Jobs for now, and may be expanded to handle other resources that will finish execution, such as Pods and custom resources.

> Alpha Disclaimer: this feature is currently alpha, and can be enabled with both kube-apiserver and kube-controller-manager feature gate TTLAfterFinished.

https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/

Please implement this maintenance feature for the following SCBv2 CRDs:
* hooks
* parser

**Additional context**

J12934:
Closes #219 

Tested locally with a kind cluster with the [TTLAfterFinished Controller](https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/) activated.


scanners/nmap/values.yaml
```diff
@@ -11,10 +11,14 @@ parserImage:
...
 
-scannerJob:
-  # scannerJob.ttlSecondsAfterFinished -- Defines how long the scanner job after finishing will be available (see: https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/)
+parseJob:
+  # parseJob.ttlSecondsAfterFinished -- seconds after which the kubernetes job for the parser will be deleted. Requires the Kubernetes TTLAfterFinished controller: https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/
   ttlSecondsAfterFinished: null
 
+scannerJob:
+# scannerJob.ttlSecondsAfterFinished -- seconds after which the kubernetes job for the scanner will be deleted. Requires the Kubernetes TTLAfterFinished controller: https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/
+  ttlSecondsAfterFinished: 0
```
rseedorff, line 20: Why is this 0 instead of null as default 🤔

scanners/ssh_scan/values.yaml
```diff
@@ -12,7 +12,7 @@ parserImage:
   tag: null
 
 scannerJob:
```
rseedorff, line 14: Is the parserJob here missing 🤔

rseedorff (changes requested):

scanners/nmap/values.yaml
J12934: Oh great find, was using nmap as my local test case and apparently screwed up changing it back to default 🤦‍♂️

scanners/ssh_scan/values.yaml
J12934: Also a screwup on my side. Fixed now

rseedorff (approved):

Issue #212, opened 2020-11-17
➹ Support ENV and File configuration options for CRDs


rseedorff:
**Is your feature request related to a problem? Please describe.**
As a secureCodeBox developer i would like to use additional extension configuration options utilizing ENV Vars, configMaps, or even secrets in the secureCodeBox CRDs. Some CRDs already had some of this already implemented and  also documented but it's not consistent.

For example this PR implements this feature for the *scan* CRD:  [Allow Specification of ENV Vars in the Scan CRD](https://github.com/secureCodeBox/secureCodeBox-v2/issues/48) 

**Describe the solution you'd like**
Please fill the gaps marked with ❌

(Note: Table updated after #231 by @J12934)

| SCBv2 CRD                                                                              | Support ENV | Documentation ENV                                                                   | Support File Mounts | Documentation File Mounts |
| -------------------------------------------------------------------------------------- | ----------- | ----------------------------------------------------------------------------------- | ------------------- | ------------------------- |
| [ScanType](https://docs.securecodebox.io/docs/api/crds/scan-type)                      | ✅          | ❌                                                                                  | ✅                  | ❌                        |
| [Scan](https://docs.securecodebox.io/docs/api/crds/scan)                               | ✅          | [✅ ](https://docs.securecodebox.io/docs/api/crds/scan#env-optional)                | ✅                  | ❌                        |
| [ParseDefinition](https://docs.securecodebox.io/docs/api/crds/parse-definition)        | ✅          | ❌                                                                                  | ✅                  | ❌                        |
| [ScanCompletionHook](https://docs.securecodebox.io/docs/api/crds/scan-completion-hook) | ✅          | [✅](https://docs.securecodebox.io/docs/api/crds/scan-completion-hook#env-optional) | ❌                  | ❌                        |

**Additional context**
- [Allow Specification of ENV Vars in the Scan CRD](https://github.com/secureCodeBox/secureCodeBox-v2/issues/48) 

**Other Todos**

- [ ] Update Ncrack Documentation to use volume mounts on the scan rather than the scan type

J12934:
Whoops should not be closed yet

rseedorff:
relates to #337

malexmave:
Related:
- https://github.com/secureCodeBox/documentation/issues/186
- https://github.com/secureCodeBox/documentation/issues/188
- https://github.com/secureCodeBox/documentation/issues/189
- https://github.com/secureCodeBox/documentation/issues/190

[...]

PR #231, opened 2020-11-26
Support for Env and File in parser and scan

paulschmelzer:
Updates the table on #212


J12934:
Looks great 🚀

This seems to be missing the ScanCompletionHook Volume & VolumeMount support? 
We can handle this in another PR, and merge this one already as it seems pretty much ready.

I've also regenerated the crds and copied them over to the operator helm chart.

J12934 (approved):

PR #243, opened 2020-12-07
Updated Amass scanner to latest release version v3.10.5

rseedorff:
## Checklist

## Description

- Updates Amass scanner to latest release version v.3.10.5
- Adds some additional amass unit tests

### DIFF
--- a/scanners/amass/parser/parser.test.js
+++ b/scanners/amass/parser/parser.test.js
@@ -1,12 +1,63 @@
-const fs = require('fs');
-const util = require('util');
+const fs = require("fs");
+const util = require("util");
 
 // eslint-disable-next-line security/detect-non-literal-fs-filename
 const readFile = util.promisify(fs.readFile);
 
-const { parse } = require('./parser');
+const { parse } = require("./parser");
 
-test('example parser parses empty json to zero findings', async () => {
+test("example parser parses empty json files to zero findings", async () => {
+  const fileContent = await readFile(__dirname + "/__testFiles__/empty.jsonl", {
+    encoding: "utf8",
+  });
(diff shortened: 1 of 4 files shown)


scanners/amass/parser/parser.test.js
```diff
@@ -1,12 +1,63 @@
...
+  expect(await parse(fileContent)).toEqual([]);
+});
+
+// test("example parser parses missing json files to zero findings", async () => {
+//   expect(await parse(null)).toEqual([]);
+// });
+
+// test("example parser parses missing json files to zero findings", async () => {
+//   expect(await parse(0)).toEqual([]);
+// });
```
fuhrmeistery, line 22: Is this meant to be commented?

fuhrmeistery (commented):
Looks good to me but there is a code snipped that is not used

scanners/amass/parser/parser.test.js
rseedorff: I introduced this test because there are some scenarios where the amass parser fails with the following error message

```
Starting Parser
Fetching result file
Fetched result file
Parser failed with error:
TypeError: fileContent.split is not a function
    at parse (/home/app/parser-wrapper/parser/parser.js:3:6)
    at main (/home/app/parser-wrapper/parser-wrapper.js:94:22)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

But i'm not sure if the parser is the best place to solve such error cases or the parser sdk @J12934

J12934: Do you still have access to the scan?
Can you take a look at the raw results? Would be interessting to see if this is just empty or what else is leading to this error.

J12934 (approved):

PR #249, opened 2020-12-18
Update diagrams

J12934:
@fuhrmeistery noticed some inconsistencies in the diagram. This should resolve this.


fuhrmeistery (commented):
Is `Notify Results` correct or should it be `Modify Results` for the `ReadAndWriteHook`

fuhrmeistery (approved):

PR #266, opened 2020-12-27
🐞 Bugfixed the image tag reference in the kube-hunter helmchart template

rseedorff:
## Checklist

## Description

The kube-hunter helmChart `scan-type` template referenced the `helmChart` Version instead of the `AppVersion`. But the scanner image docker tag references to the AppVersion as one can see here: https://hub.docker.com/repository/docker/securecodebox/scanner-kube-hunter/tags?page=1&ordering=last_updated

Maybe it would be easier to be more consistent here 🤔 Some of the scanner image docker tags are referencing the SCB release version and some are referencing the underling app version of the security scanner used.

### DIFF
--- a/scanners/kube-hunter/templates/kubehunter-scan-type.yaml
+++ b/scanners/kube-hunter/templates/kubehunter-scan-type.yaml
@@ -17,7 +17,7 @@ spec:
           restartPolicy: Never
           containers:
             - name: kube-hunter
-              image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.Version }}"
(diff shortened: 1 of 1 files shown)


fuhrmeistery (approved):

J12934:
All third party scanners should have their own version number.
Our own "scanners" (git-repo-scan, screenshooter & test-scan) use the secureCodeBox version numbers as they are part of the secureCodeBox. We could also make them have their own version numbers but this would require extra effort.

@rseedorff Which charts did you notice where the secureCodeBox version was used, just the three? Or did we screw up more than the kube-hunter one?

PR #300, opened 2021-02-17
Add Hook to Import Scan Results into DefectDojo

J12934:
This PR adds a new Hook which imports raw results of scans into DefectDojo.
For more details check out the hooks readme: https://github.com/secureCodeBox/secureCodeBox/blob/feature/defect-dojo/hooks/persistence-defectdojo/README.md

You can test out this pr by using the official DefectDojo Demo instance at: https://demo.defectdojo.org/
See DefectDojo docs for the credentials: https://defectdojo.readthedocs.io/en/latest/

To test out the pr with the demo instance, you can install the hook like the following:

```bash
# check out secureCodeBox repo at feature/defect-dojo branch

# log in to defectdojo beforehand and grab the apiv2 key for your user. Then replace the apikey placeholder in the command below
kubectl create secret generic defectdojo-credentials --from-literal="username=admin" --from-literal="apikey=08b7..."

helm upgrade --install defectdojo-hook ./hooks/persistence-defectdojo/ \
  --set="defectdojo.url=https://demo.defectdojo.org" \
  --set="image.tag=feature-defect-dojo"
```


J12934:
Ok, I got to say I'm a bit confused with code climate here.
The one method it doesn't like (`getProductName`) is probably one of the best readable ones in the PR.
Does anybody have an idea on how to improve it?

damiencarol:
@J12934 I think it's because this method call a method of a proxy object. You should just remove it use scan.getProductName() is your code.
```
  protected String getProductName(Scan scan) {
    if (scan.getProductName().isPresent()) {
      return scan.getProductName().get();
    }
```
Also 5 in complexity is a little bit low :D

hooks/persistence-defectdojo/build.gradle
```diff
@@ -0,0 +1,50 @@
...
+  jcenter()
+  maven {
+    url = "https://oss.sonatype.org/content/repositories/snapshots/"
+  }
+}
+
+dependencies {
+  implementation 'io.kubernetes:client-java:10.0.0'
+
+  implementation 'io.securecodebox:defectdojo-client:0.0.7-SNAPSHOT'
```
rseedorff, line 21: We reference here a released version instead of a snapshot version.

J12934: Yes 👍
I planned to do this later after this initial merge once we're fairly certain that the API of the client is stable.

hooks/persistence-defectdojo/templates/_helpers.tpl
```diff
@@ -0,0 +1,52 @@
+{{/* vim: set filetype=mustache: */}}
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "persistence-elastic.name" -}}
```
rseedorff, line 5: elastic != defectdojo

damiencarol:
@J12934 I saw that it's merged now! :+1: 
Do we have a demo env to test this one?

PR #305, opened 2021-02-25
Move to seperate Kubernetes ServiceAccount for the Operator

J12934:
Previous to this PR the secureCodeBox Operator was using the default ServiceAccount in the `securecodebox-system` namespace. This is the default for kubebuilder projects but never really felt right, as this makes it harder to track and might cause conflicts with different secureCodeBox related services running in the `securecodebox-system` namespace (😉).

I've had this problem while integrating the secureCodeBox Operator with AWS IAM Role to Kubernetes ServiceAccount binding (IRSA), see: https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/
In this mechanism the Kubernetes ServiceAccount gets linked to a AWS IAM Role via a annotation on the serviceAccount which then makes the AWS EKS Cluster inject a AWS STS Token into the pod automatically. 

So in short, this pr makes the operator run under it's own serviceAccount `securecodebox-operator` in the `securecodebox-system` namespace. And also adds some missing helm values required to properly setup AWS IRSA Authentication with the operator, will add a section on that in the installation guide.


operator/Dockerfile
```diff
@@ -21,7 +21,10 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:nonroot
```
fuhrmeistery: It should also be possible to debug distroless images by using `debug` as tag or if you already use a tag by appending `-debug`. See https://github.com/GoogleContainerTools/distroless#debug-images

fuhrmeistery (approved):

operator/Dockerfile
J12934: Not for the `:nonroot` images 🙁
And images with root would not have been possible to run with the default security context 🙁

PR #315, opened 2021-03-07
Update SCB Scanner dependencies

rseedorff:
## Checklist

## Description

This PR if applied updates the SCB Scanner Versions to the latest major release.

### DIFF
--- a/scanners/wpscan/Chart.yaml
+++ b/scanners/wpscan/Chart.yaml
@@ -5,7 +5,7 @@ description: A Helm chart for the WordPress security scanner that integrates wit
 type: application
 # version - gets automatically set to the secureCodeBox release version when the helm charts gets published
 version: v2.5.0-alpha1
-appVersion: latest
+appVersion: v3.8.15
 kubeVersion: ">=v1.11.0-0"
 
 keywords:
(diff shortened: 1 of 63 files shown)


scanners/wpscan/Chart.yaml
```diff
@@ -4,8 +4,8 @@ description: A Helm chart for the WordPress security scanner that integrates wit
 
 type: application
 # version - gets automatically set to the secureCodeBox release version when the helm charts gets published
-version: v2.5.0-alpha1
-appVersion: latest
+version: v2.6.0-alpha1
+appVersion: v3.8.15
```
J12934, line 8: We could pin the image to the current digest, otherwise the 3.8.15 will be wrong after the next wpscan update.

rseedorff: Mhh yeah, but the problem is that there is only a latest tag available an no versioning schema https://hub.docker.com/r/wpscanteam/wpscan/tags?page=1&ordering=last_updated 👎 

I guess pinning to a specific digest wouldn't solve this problem :-/

I'll revert this change here back to latest and hopefully the wpscan will fix that in the feature an support tags with release versions.

J12934 (approved):

Issue #325, opened 2021-03-12
🐞 The git_repo_scanner triggers an abuse detection mechanism at GitHub for large organisations 


glermaidt:
**Describe the bug**

I am having an issue with a git-repo-scanner + gitleaks working in concert to scan my company's org, which has over 2k repos.  The scan is ending around 900 repos with the following error:

```
INFO:git_repo_scanner: 901 - crp-legacy-ui-dhtmlWins
INFO:git_repo_scanner: Github API Exception: 403 -> You have triggered an abuse detection mechanism. Please wait a few minutes before you try again.
```

This is causing CrashLoopBackoff events with the pod.  Ultimately the job fails with error.

I understand what the error is telling me.  My main question is:  What is the largest number of repos does this scanner support? I would like to scan my entire org in one run, but I have not been successful, with the number of repos and the GH API throttle limits.  The --ignore-repo is a start, but seems not too effective given my large list of repos.

Other questions:
Is there a way to throttle api calls? 
Is there a recommended number of batches I should run with?
Would there be a way to pass an argument to break up the scans?

**To Reproduce**

1. Install via helm:  see second screenshot below.
2. Create a git-repo-scanner yaml file.
3. Run the git-repo-scanner job on a github org with over 2000 repos and with cascadingrules triggering gitleaks.
4. Tail the created job's log, for example: `k logs job/scan-gh-repo-scan-org-coretech-wdbns -c git-repo-scanner -f `
4. See 403 error in the first screenshot below.

**Expected behavior**

**System (please complete the following information):**

 - secureCodeBox 2.4.0
 - OS: Linux
 - Kubernetes Version [command: `kubectl version`]
 
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.4", GitCommit:"e87da0bd6e03ec3fea7933c4b5263d151aafd07c", GitTreeState:"clean", BuildDate:"2021-02-21T20:23:45Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.9-eks-d1db3c", GitCommit:"d1db3c46e55f95d6a7d3e5578689371318f95ff9", GitTreeState:"clean", BuildDate:"2020-10-20T22:18:07Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
 - Docker Version [command: `docker -v`]
 
[...]

Issue #320, opened 2021-03-09
⚙️ Extend the git-repo-scanner with an activity filter


rseedorff:
**Is your feature request related to a problem? Please describe.**

As secureCodeBox user i am heavily using the `git-repo-scanner` in combination with the `gitleaks` scanner via cascadingRules. When scanning large Gitlab oder GitHub organisations _with thousands of repos_ on a regular basis (daily, weekly) this leeds to thousands of independent `gitleaks` scans. 

To save resources on the `gitleaks` side we extended `gitleaks` to be able to analyse all commits since a given timeframe (last 24h, last week,...): https://github.com/zricethezav/gitleaks/pull/498

Problem is that the `git-repo-scanner` always returns all git repositories, even if they had no activity in the last timeframe. To safe even more resources it would be great to be able to configure an activity timeframe and use it as filter.

**Describe the solution you'd like**

Add the following configuration options to the `git-repo-scanner`:

```
--activity-since-duration= Return git repo findings with repo activity (e.g. commits) more recent than a specific date expresed by an duration (now + duration). A duration string is a possibly signed sequence of decimal numbers, each
                               with optional fraction and a unit suffix, such as '300ms', '-1.5h' or '2h45m'. Valid time units are 'ns', 'us' (or 'µs'), 'ms', 's', 'm', 'h'.
--activity-until-duration= Return git repo findings with repo activity (e.g. commits) older than a specific date expresed by an duration (now + duration). A duration string is a possibly signed sequence of decimal numbers, each with
                               optional fraction and a unit suffix, such as '300ms', '-1.5h' or '2h45m'. Valid time units are 'ns', 'us' (or 'µs'), 'ms', 's', 'm', 'h'.
```
_Remark 🚧  : The concrete semantic of the timeframe definition can depend on the implementation details and might be slightly different. It is not important to implement it in exact that semantic._

### Implementation Hints:
- The GitLab Python client has some helpful methods like sorting all projects by `last_activity_at`: https://python-gitlab.readthedocs.io/en/stable/gl_objects/projects.html#examples 
  - `gl.projects.list(all=True, include_subgroups=True, order_by='last_activity_at', sort='desc')`
  - could be used at: https://github.com/secureCodeBox/secureCodeBox/blob/fab01bdff212dde428e4dc9eb6eae4f34e211a98/scanners/git-repo-scanner/scanner/git_repo_scanner.py#L131
[...]

PR #327, opened 2021-03-13
Added new git-repo-scanner options (rate limit, activity filter)

rseedorff:
### Implements a new rate limit configuration
Implements #325 

This PR if applied adds a new rate limit configuration to the git-repo-scanner to obey the GitHub and GitLeaks server rate limits. By default this setting is true.

#### Example
Works for **GitHub** and **Gitlab**:
```bash
--obey-rate-limit: True to obey the rate limit of the GitHub server (default), otherwise False
```

* For unauthenticated requests, the rate limit allows for up to 60 requests per hour. Unauthenticated requests are associated with the originating IP address, and not the user making requests. 
  * With `--obey-rate-limit=True` the scanner will hit the GitHub API every 60 seconds at max (_3600s / 60 requests per hour = 60s seconds delay between two request_).
* OAuth applications authorized by a user share the same quota of 5,000 requests per hour when they authenticate with different tokens owned by the same user. 
  * With `--obey-rate-limit=True` the scanner will hit the GitHub API every 0,7 seconds at max (_3600s / 5000 requests per hour = 0,7s seconds delay between two request_).

### Implements a new git repository filter based on latest activity
Implements #320 

#### Example
Works for **GitHub** and **Gitlab**:

```bash
--activity-since-duration: Return git repo findings with repo activity (e.g. commits) more recent than a specific date expresed by an duration (now + duration). A duration string is a possibly signed sequence of decimal numbers, each
                               with optional fraction and a unit suffix, such as '1h' or '2h45m'. Valid time units are 'm', 'h', 'd', 'w'.
--activity-until-duration: Return git repo findings with repo activity (e.g. commits) older than a specific date expresed by an duration (now + duration). A duration string is a possibly signed sequence of decimal numbers, each with
                               optional fraction and a unit suffix, such as '1h' or '2h45m'. Valid time units are 'm', 'h', 'd', 'w'.
```

### Testing
If you want to test this PR please reference the following docker image: `docker pull securecodebox/scanner-git-repo-scanner:git-repo-rate-limit`

```yaml
image:
  # image.repository -- Container Image to run the scan
  repository: docker.io/securecodebox/scanner-git-repo-scanner
  # image.tag -- defaults to the charts version
  tag: git-repo-rate-limit
```

#### Example HelmChart Install
You can use the following command to install and test this PR in your environment. For more general informations please check the scanner documentation : https://docs.securecodebox.io/docs/scanners/git-repo-scanner
```bash
# Optional: Add the secureCodeBox Helm Repo
helm repo add secureCodeBox https://charts.securecodebox.io
# Install HelmChart (use -n to configure another namespace)
helm upgrade --install git-repo-scanner secureCodeBox/git-repo-scanner \
  --set="image.repository=docker.io/securecodebox/scanner-git-repo-scanner" \
  --set="image.tag=git-repo-rate-limit"
```
## Checklist

* [ ] Cleanup the code to make codeclimate happy


rseedorff:
This PR already works (on my machine 😆) as expected 🎉 

But adding the functionality in a hurry 🏃🏻‍♂️ results in a bit messy python code 🙄 The codebase must be refactored and we should introduce a more appropriate python class based implementation:
- GitHub Class
- GitLab Class

This would also introduce new possibilities to integrate even more features in the `git-repo-scanner` like a signed commit checker for example 😉

PR #374, opened 2021-04-14
Allow wildcards in CascadingRules (closes #357)

### LINKED ISSUE #357: Allow wildcards in the CascadingRules

As a secureCodeBox user I want tot use the CascadingRules to start scans based on the results of other scans. Currently every single rule only checks if the service name is an exact match which can result in a long list of rules.

Allowing wildcards would allow users to define one CascadingRule that would work for multiple services, for instance `ssl/*` to start a scan for multiple types of SSL services.

Defining every possible service in a rules file which would be a lot of manual work for every secureCodeBox user editing rules.

luckolen:
## Description

This PR add the option for users to use the * as a wildcard when defining CascadingRules. For instance `http*` will match everything starting with `http` and as a result one rule can do both `http` and `https`.

### Checklist


hooks/declarative-subsequent-scans/hook.ts
```diff
@@ -127,3 +129,16 @@ function generateCascadingScanName(
...
   return `${namePrefix}-${cascadingRule.metadata.name}`;
 }
+
+function wildcardMatcher(
+  findingValue: any,
+  matchesRuleValue: any
+) : boolean {
+  if(isString(matchesRuleValue)){
+    try{
+      return new RegExp('^' + new String(matchesRuleValue).replace(/\*/g, '.*') + '$').test(findingValue);
```
J12934, line 139: I think it might be worth it to delegate the wildcard matching to a library, as this is quite hard to get completely right, for example the might be cases where you'd want to escape the `*` and have it match the character not as a wildcard.

I think a good lib for it would be the `isMatch` function from https://www.npmjs.com/package/matcher

luckolen: Thank you for suggesting that lib, this has been implemented in https://github.com/secureCodeBox/secureCodeBox/pull/374/commits/5884c89ede46c33716bec800f3482c3feefeefcb which is included in this PR.

I've noticed that there is a parameter to make the matching case sensitive. I've set it to true to match current application behavior, but I can see a case for having the CascadingRules be case-insensitive.

J12934: Awesome thank you 👍

PR #424, opened 2021-05-17
Fix Crash in Wpscan Parser When No Wordpress Version Got Identified

J12934:
This PR fixes a issue in the WPScan Parser which causes it to crash when WPScan wasn't able to identify the WPScan version.

Also upgraded the ecmascript version specified in eslint to match our node.js version used so that we are can use modern js features like optional chaining without eslint complaining about it.


rseedorff (approved):

rseedorff:
Great catch @J12934 kudos for fixing and improving that 👍🏻. Like your new implementation, that’s a lot cleaner solution 😊

PR #425, opened 2021-05-17
Improve install/uninstall script to respect namespaces

rseedorff:
## Description

This PR if applied improves the `bin/install.sh` and `bin/uninstall.sh` scripts to respect namespaces. This is helpful for different installation scenarios.

### Checklist


bin/install.sh
```diff
@@ -210,19 +226,22 @@ function interactiveInstall() {
...
+    installResources "$BASE_DIR/demo-apps" "$SCB_DEMO_NAMESPACE" True
   fi
 
-  if [[ -n "${INSTALL_DEMO_APPS}" ]]; then
-    print "Starting to install demo-apps..."
-    installResources "$BASE_DIR/demo-apps" "default" True
+  if [[ -n "${INSTALL_SCANNERS}" ]]; then
+    print "Starting to install 'scanners' into namespace '$SCB_NAMESPACE' ..."
+    kubectl create namespace "$SCB_NAMESPACE" || print "Namespace '$SCB_NAMESPACE' already exists or could not be created.. "
+    installResources "$BASE_DIR/scanners" "$SCB_NAMESPACE" True
```
Weltraumschaf, line 238: Using `True` in Bash is completely wrong! There is no such boolean type in Bash. The problem is: This is simply a string `"True"`, nothing more (not to confuse w/ `true` which is the binary `/usr/bin/true` which simply always exits with 0). So checking `$foo=True` may lead to errors in various ways: This will not match if $foo is "true". It will fail if there is a binary named "True" in the PATH of the executing shell and so on.

So the pattern to circumvent missing booleans in Bash is as followed:

- We agree on the assumption that boolean false is always represented as empty string (`''`).
- We agress on the assumption that boolean true is always represented as **non** empty string. By convention we use the string "true", but in fact it does not really matter. It's only to not confusing others.
- We always check for (non) emptyness. E.g. `[[ -n "${INSTALL_DEMO_APPS}" ]]` (evaluates to true if `$INSTALL_DEMO_APPS` is not empty. To test the other way use `-z`.

You can see examples of this pattern e.g. for the CLI flag `--demo-apps`.

bin/install.sh
```diff
@@ -210,19 +226,22 @@ function interactiveInstall() {
...
   fi
 
   if [[ -n "${INSTALL_HOOKS}" ]]; then
-    print "Starting to install hooks..."
-    installResources "$BASE_DIR/hooks" "default" True
+    print "Starting to install 'hooks' into namespace '$SCB_NAMESPACE' ..."
+    kubectl create namespace "$SCB_NAMESPACE" || print "Namespace '$SCB_NAMESPACE' already exists or could not be created.. "
+    installResources "$BASE_DIR/hooks" "$SCB_NAMESPACE" True
```
Weltraumschaf, line 244: See above.

bin/uninstall.sh
```diff
@@ -28,14 +33,14 @@ function uninstallResources() {
 
   for resource in "${resources[@]}"; do
     local resource_name="${resource//+([_])/-}" # Necessary because ssh_scan is called ssh-scan
-    helm uninstall "$resource_name" || true
+    helm uninstall "$resource_name" -n $namespace || true
```
Weltraumschaf, line 36: You should always quote the variable to prevent unwanted word splitting.

bin/uninstall.sh
```diff
@@ -28,14 +33,14 @@ function uninstallResources() {
...
   done
 }
 
-helm -n securecodebox-system uninstall securecodebox-operator || true
+helm -n $SCB_SYSTEM_NAMESPACE uninstall securecodebox-operator || true
```
Weltraumschaf, line 40: You should always quote the variable to prevent unwanted word splitting.

bin/uninstall.sh
```diff
@@ -28,14 +33,14 @@ function uninstallResources() {
...
 
-uninstallResources "$BASE_DIR/scanners"
-uninstallResources "$BASE_DIR/demo-apps"
-uninstallResources "$BASE_DIR/hooks"
+uninstallResources "$BASE_DIR/demo-apps" "$SCB_DEMO_NAMESPACE"
+uninstallResources "$BASE_DIR/scanners" "$SCB_NAMESPACE"
+uninstallResources "$BASE_DIR/hooks" "$SCB_NAMESPACE"
 
-kubectl delete namespaces securecodebox-system || true
+kubectl delete namespaces $SCB_SYSTEM_NAMESPACE || true
```
Weltraumschaf, line 46: You should always quote the variable to prevent unwanted word splitting.

Weltraumschaf (changes requested):

bin/install.sh
```diff
@@ -100,19 +105,19 @@ function exitIfHelmIsNotInstalled() {
...
-  kubectl create namespace securecodebox-system || print "Namespace already exists..."
+  print "Creating namespace $SCB_SYSTEM_NAMESPACE"
+  kubectl create namespace $SCB_SYSTEM_NAMESPACE || print "Namespace '$SCB_SYSTEM_NAMESPACE' already exists..."
 
-  print "Installing the operator in the securecodebox-system namespace"
+  print "Installing the operator in the '$SCB_SYSTEM_NAMESPACE' namespace"
 
-  if [[ $(helm -n securecodebox-system upgrade --install securecodebox-operator "$BASE_DIR"/operator/) ]]; then
-    print "$COLOR_OK" "Successfully installed the operator!"
+  if [[ $(helm -n $SCB_SYSTEM_NAMESPACE upgrade --install securecodebox-operator "$BASE_DIR"/operator/) ]]; then
```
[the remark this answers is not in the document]
rseedorff: good point! thx

bin/install.sh
rseedorff: Thx for your explanation, didn't know that point! 

But i'm a bit confused here because the code block you mentioned here is exactly the same as your example: `[[ -n "${INSTALL_DEMO_APPS}" ]]` =  `if [[ -n "${INSTALL_SCANNERS}" ]]; then` 🤔

We can agree on the assumption but the result (codeblock) will be the same als already is, or did i miss something here?

Weltraumschaf: The point here for this particular snippet is not the test itself, but the content of the tested variable. You assigned `True` which is accidentally converted to the String `"True"` and so this works by accident. This pattern in Bash consist of two main parts:

1. The value used to represent a boolean (empty vs non-empty string).
2. The test of this value with -n or -z.

The second part is done right here. But the first one was not.

bin/install.sh
```diff
@@ -37,6 +37,11 @@ INSTALL_INTERACTIVE=''
 INSTALL_SCANNERS=''
 INSTALL_DEMO_APPS=''
 INSTALL_HOOKS=''
+INSTALL_NAMESPACED='false'
```
Weltraumschaf, line 40: Here you should use empty string to indicate false.

bin/install.sh
```diff
@@ -192,14 +193,29 @@ function interactiveInstall() {
...
+  if [[ $line == *[Yy] ]]; then
+    print "Please provide a name for the namespace:"
+    read -r NAMESPACE
+    kubectl create namespace "$NAMESPACE" || print "Namespace '$NAMESPACE' already exists or could not be created!"
+  fi
+
   print
   print "Starting to install hooks..."
-  installResources "$BASE_DIR/hooks" "default" False
+  installResources "$BASE_DIR/hooks" "$NAMESPACE" False
```
Weltraumschaf, line 214: Here you should use just an empty string or nothing to indicate false.

bin/install.sh
```diff
@@ -192,14 +193,29 @@ function interactiveInstall() {
...
+
+  print
+  print "Starting to install scanners..."
+  installResources "$BASE_DIR/scanners" "$NAMESPACE" False
```
Weltraumschaf, line 218: Same here.

bin/install.sh
```diff
@@ -214,19 +230,22 @@ function interactiveInstall() {
 }
 
 function unattendedInstall() {
-  if [[ -n "${INSTALL_SCANNERS}" ]]; then
-    print "Starting to install scanners..."
-    installResources "$BASE_DIR/scanners" "default" True
+  if [[ -n "${INSTALL_DEMO_APPS}" ]]; then
+    print "Starting to install 'demo-apps' into namespace '$SCB_DEMO_NAMESPACE' ..."
+    kubectl create namespace "$SCB_DEMO_NAMESPACE" || print "Namespace '$SCB_DEMO_NAMESPACE' already exists or could not be created!"
+    installResources "$BASE_DIR/demo-apps" "$SCB_DEMO_NAMESPACE" True
```
Weltraumschaf, line 236: Here you should use `"true"` (as string) to indicate true.

bin/install.sh
Weltraumschaf: I commented the places above: Where you use literal `True` or `False` you should use `"true"` or `""`.

Weltraumschaf (changes requested):

Weltraumschaf (approved):
LGTM

PR #437, opened 2021-05-20
Defectdojo generic parser integration

### LINKED ISSUE #432: Enable compatibility with DefectDojo parsers that rely on file names
The DefectDojo Nikto Scanner can currently not be used because it relies on file names or more precisely their endings like ".xml" or ".json".

DefectDojo Parsers that rely on file names can be used

### Additional context
This requires changes in the  public persistence-defectdojo hook in the securecodebox and also in the private defectdojo-java-client

JohannesZahn:
## Description

If no explicit parser is specified for a scan, a Generic Parser will be selected and instead of the raw scan results the generic findings.json will be sent to DefectDojo. 

This will currently not work for the generic case, because the SecureCodeBox Parser in DefectDojo is not yet released and also it is still up for debate if the generic JSON Parser should be used instead.

Still it would be beneficial to merge the PR as it enables both approaches and the code does not work any less than now, as it still works for the parsers that are already implemented. Now however the logic for selecting the generic parser and downloading the findings.json enables the second feature introduced by this PR:

File names (for the generic as well as specific parsers) are now sent to DefectDojo, so parsers that rely on filenames like nikto, as well as the generic json parser can now be selected. This therefore closes #432. 

### Checklist

### DIFF
--- a/hooks/persistence-defectdojo/src/main/java/io/securecodebox/persistence/DefectDojoPersistenceProvider.java
+++ b/hooks/persistence-defectdojo/src/main/java/io/securecodebox/persistence/DefectDojoPersistenceProvider.java
@@ -26,7 +26,6 @@ import io.securecodebox.persistence.service.KubernetesService;
 import io.securecodebox.persistence.service.S3Service;
 import io.securecodebox.persistence.strategies.VersionedEngagementsStrategy;
 import io.securecodebox.persistence.util.ScanNameMapping;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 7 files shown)


hooks/persistence-defectdojo/src/main/java/io/securecodebox/persistence/DefectDojoPersistenceProvider.java
```diff
@@ -46,17 +49,27 @@ public static void main(String[] args) throws Exception {
...
+    var scanResults = s3Service.downloadFile(downloadUrl);
+    LOG.info("Finished Downloading Scan Result");
+    LOG.debug("Scan Result: {}", scanResults);
 
     LOG.info("Uploading Findings to DefectDojo at: {}", config.getUrl());
 
     var defectdojoImportStrategy = new VersionedEngagementsStrategy();
     defectdojoImportStrategy.init(config);
-    var defectDojoFindings = defectdojoImportStrategy.run(scan, rawResults);
+    var scanResultFile = new ScanFile(scanResults, FilenameUtils.getName(new URL(downloadUrl).getPath()));
```
J12934, line 71: You should be able to get the filename of the raw results file directly from the Scan status, instead of parsing it from the downloadUrl:

```suggestion
    var scanResultFile = new ScanFile(scanResults, scan.getStatus().getRawResultFile());
```

J12934 (changes requested):

J12934 (approved):

PR #441, opened 2021-05-21
Adds the Developer Certificate of Origin and Contributors

Weltraumschaf:
Adds the Developer Certificate of Origin and Contributors.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -1,0 +1,19 @@
+# Contributors
+
+If you would like to make contributions to this repository, please certify to the following Developer's Certificate of Origin (DCO) in [this repository](DCO). Please confirm your certification to the DCO by adding a sign-of tag to all your commit messages and adding your author name (e.g. `gordon shumway <alf@melmak.com>`) to the following list of contributors and open a pull request.
+
+If you are a developer who is authorized to contribute on behalf of your employer, then please use your corporate email address in the signed-off-by tag. Unless please use your personal email address.
+
+## Sign-of Tag
+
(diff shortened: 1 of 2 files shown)


rseedorff (approved):

CONTRIBUTORS.md
```diff
@@ -0,0 +1,19 @@
...
+
+```text
+Signed-off-by: gordon shumway <alf@melmak.com>
+```
+
+Committing with `git commit -s` will add the sign-off at the end of the commit message automatically for you.
+
+## List of Contributors
+
+- Sven Strittmatter <sven.strittmatter@iteratec.com>
```
fuhrmeistery, line 19: Maybe it would make sense to discriminate between contributor and maintainer?

Weltraumschaf: Have have no idea :-)

Didn't found any example for that. In the first place adding your self here is like "signing the fax" in the CLA process. So, IMHO we need not to mind it here at the moment.

fuhrmeistery: I don't remember the repo that I had open at this point, but Ansible has separate files for maintainers and contributors 
See https://github.com/ansible-collections/community.postgresql

J12934: I don't understand what this list is for?
Is this a "Hall of Fame" style of list, or is this required for the DCO part?

Weltraumschaf: They just use plain text files with the GH handle and they do not use DCO anyway. So I'm not convinced why this is useful.

I suggest to move this discussion into the related issue and merge this so we can sign the DCO.

fuhrmeistery (approved):

PR #469, opened 2021-06-03
Add `scanner.appendName` to chart values

EndPositive:
## Description
Using {{ .Release.name }} causes issues when using this chart as a dependency of another chart. All other scanners already use a fixed name.

### Checklist


EndPositive:
I believe no tests/examples are making use of the release name "feature". Everywhere I can see, this `ScanType` is referenced as `nmap`.

J12934:
Good catch, I agree that this should be changed at least so that it does not use the helm release name as the name of the scanType as this really isn't intuitive. I think we should hold this change untill 3.0.0 as this is a potentially breaking change.

This was originally added so that you can configure yourself a `nmap-privilidged` scanType, which is capable of running operating system scans which requires some higher privileges: https://docs.securecodebox.io/docs/scanners/nmap#operating-system-scans

In v3 we should then probably add a helm value to control the name of the scanTypes and not use the release name

EndPositive:
Interesting, I hadn't thought of that. I agree that a helm value for overriding the scantype name would be a great option for that use case. I was assuming we could do with a single override value for each chart (i.e. `scannerJob.name`), but that wouldn't work out for charts with multiple scantypes defined (e.g. zap).

1. A simple solution would be to make a value for appending to the existing names, such that if `scannerJob.nameAppend`: `-privileged` would create `zap-baseline-scan-privileged`, `zap-api-scan-privileged`, `zap-full-scan-privileged`. 

2. Another solution would be to allow users to manually override the names of each scantype with `scannerJob.zap-baseline-scan`:`zap-baseline-scan-privileged`, `scannerJob.zap-api-scan`:`zap-api-scan-privileged`, etc. 

I'm not sure if we have a use case for option 2 since each scantype also already uses the same security, resources, etc.

J12934:
Hi, sorry this pr got out of my sight. 😕
I really like the first solution you proposed, that sounds like a great way to add this while allowing all charts to keep a consistent value schema 👍

EndPositive:
I think that should do it :smile:

J12934 (approved):

J12934:
Commit 4ec547e is missing the DCO sign-off 🥴

EndPositive:
I can't figure out how to update the commit without messing up the commit history. All the commits from this merge 75c9727c5cbe781410c68f24db2a5c265ea5ecc5 are also rebased... and then the commits show up duplicated in the logs.

If you want I can create a new PR? Or just merge it, since the DCO was only introduced after 4ec547e30908946327e64eecc9aec076bec73686

Issue #476, opened 2021-06-08
Makefiles For CI And Local Testing


fuhrmeistery:
As developer/maintainer of the *secureCodeBox*
I want to reduce the complexity of installing unreleased scanners/hooks
so that testing locally and in the CI becomes more easy.

To tackle this issue, we want to introduce Makefiles for each scanner/hook.
- [ ] ~~Add Makefiles for all Hooks #478~~
- [ ] ~~Add Makefiles for all Scanners #477~~
- [ ] #622

PR #494, opened 2021-06-14
Introduce Testing Makefiles For Amass Scanner And Finding Post Processing Hook

### LINKED ISSUE #495: Align Hook Directory Structure With Scanner
I noticed that scanners and hooks use a different approach in their directory structure.
For the scanners there are subdirectories for `parser` and `scanner` (if existing) while the source code for the hook resides directly in its root directory.

I think we should introduce a `hook` directory inside each hook containing its source code to make their structure more clean.

fuhrmeistery:
- Add Initial Test Makefile for scanner/amass
- Add Initial Test Makefile for hook/finding-post-processig
- Closes #495 

See: #476


scanners/amass/parser/e2e/amass.test.js
```diff
@@ -0,0 +1,21 @@
+// SPDX-FileCopyrightText: 2020 iteratec GmbH
```
J12934, line 1: I think the e2e directory should be directly located in the `amass/` directory not in the `amass/parser/` directory as the point of a e2e test is to test it all together not just the parser.

scanners/amass/parser/e2e/helpers.js
```diff
@@ -0,0 +1,280 @@
+// SPDX-FileCopyrightText: 2020 iteratec GmbH
```
J12934, line 1: The helper file is just here temporarily for testing purposes?
Duplicating it for every scanner & hook sounds tedious.

scanners/amass/parser/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "secureCodeBox"
+  ],
+  "author": "iteratec GmbH",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/securecodebox/securecodebox/issues"
+  },
+  "homepage": "https://github.com/securecodebox/securecodebox#readme",
+  "devDependencies": {
+    "@kubernetes/client-node": "^0.14.3",
```
J12934, line 25: From a maintainer perspective it would be awesome if we could find a way to run the tests without having every scanner bring their own package.json & jest installation as this will cause dependabot to open up dozens of PR's on new vulnerabilities.

Maybe we could use `npx jest` / `npx -p jest@27 jest` from the makefiles? Would probably need to have kubernetes client-node installed either on the top level in the repo / globally to work.

hooks/finding-post-processing/Makefile
```diff
@@ -0,0 +1,76 @@
...
+
+build-images:
+	@echo Build With BASE_IMG_TAG: $(BASE_IMG_TAG)
+	docker build --build-arg=baseImageTag=$(BASE_IMG_TAG) --build-arg=namespace=$(IMG_NS) -t $(IMG_NS)/$(hook-prefix)$(hook):$(IMG_TAG) .
+
+images-export:
+	docker save $(IMG_NS)/$(hook-prefix)$(hook):$(IMG_TAG) -o $(hook-prefix)$(hook).tar
+
+kind-import:
+	kind load image-archive ./$(hook-prefix)$(hook).tar
```
J12934, line 59: These look like they would be exactly the same for every scanner, might be easier from a maintance perspective to have them in the CI matrix script instead of duplicating them in the makefile.

hooks/finding-post-processing/Makefile
```diff
@@ -0,0 +1,76 @@
...
+		--values ./e2e/__testFiles__/values.yaml \
+		--set="image.repository=docker.io/$(IMG_NS)/$(hook-prefix)$(hook)" \
+		--set="image.tag=$(IMG_TAG)"
+
+deploy-test-deps:
+	helm -n integration-tests upgrade --install test-scan ../../scanners/test-scan/
+
+integration-tests:
+	kubectl -n integration-tests delete scans --all
+	npm test --ci --color ./e2e/finding-post-processing.test.js
```
J12934, line 72: Would love it if we could agree with this change whether to call the tests `e2e` or `integration` test and use the wording consistently 😅

hooks/finding-post-processing/Makefile
fuhrmeistery: Yes and No. For most of the scanners, this is equal, with an exception to the scanners we provide custom Images for. So yes it could be managed with matrix builds. 
My original idea behind these steps was to simplify the local testing.

scanners/amass/parser/e2e/helpers.js
fuhrmeistery: Just for easy decoupled testing :+1:

hooks/finding-post-processing/Makefile
J12934: Ah didn't think of the "custom build scanners" 👍

scanners/amass/parser/package.json
fuhrmeistery: The Problem is that some parsers/hooks like finding-post-processing have their own dependencies.

hooks/finding-post-processing/.helmignore
```diff
@@ -32,4 +32,5 @@ src/*
 config/*
 Dockerfile
 .dockerignore
-
+finding-post-processing.tar
```
rseedorff, line 35:
```suggestion
*.tar
```
I would recommend to exclude all kinds of tar files in all `.helmignore` files instead of an specific exclude

rseedorff (changes requested):

tests/integration/scanner/amass.test.js
```diff
@@ -2,7 +2,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-const { scan } = require("../helpers");
```
J12934, line 5: shouldn't this file be deleted?

scanners/amass/.gitignore
```diff
@@ -0,0 +1 @@
+parser-amass.tar
```
J12934, line 1: Could we add a general ignore all tars to the top level gitignore?

J12934:
The PR title is now pretty misleading as this does very little to scanners and ver much for hooks 😜

hooks/finding-post-processing/Makefile
```diff
@@ -44,45 +50,66 @@ hook-prefix =
...
+	# If not exists create namespace where the tests will be executed
+	kubectl create namespace integration-tests --dry-run=client -o yaml | kubectl apply -f -
+	# If not exists create secureCodeBox operator namespace
+	kubectl create namespace securecodebox-system --dry-run=client -o yaml | kubectl apply -f -
+	helm -n securecodebox-system upgrade --install securecodebox-operator secureCodeBox/operator --wait \
+		# --set="image.repository=docker.io/$(IMG_NS)/operator" \
+		# --set="image.tag=$(BASE_IMG_TAG)" \
+		# --set="lurcher.image.repository=docker.io/$(IMG_NS)/lurcher" \
+		# --set="lurcher.image.tag=$(BASE_IMG_TAG)" \
+
```
fuhrmeistery: I am not sure if we should install the operator like this for 3 reasons:
1. You are forced to rebuild for every commit when trying to execute the Makefile locally (because operator and scanner/hook use the same image tag pinned to the last commit)
2. This will probably crash when you run the operator locally
3. Changes in the operator (its values) could affect all Hook Scanner Makefiles in the future

I would propose to outsource this step into the Makefile for the operator

scanners/amass/Makefile
```diff
@@ -45,42 +51,61 @@ parser-prefix = parser
...
+	@echo ".: 💾 Deploying prerequisite HelmChart with the docker tag '$(BASE_IMG_TAG)' into kind."
+	# If not exists create namespace where the tests will be executed
+	kubectl create namespace integration-tests --dry-run=client -o yaml | kubectl apply -f -
+	# If not exists create secureCodeBox operator namespace
+	kubectl create namespace securecodebox-system --dry-run=client -o yaml | kubectl apply -f -
+	helm -n securecodebox-system upgrade --install securecodebox-operator secureCodeBox/operator --wait \
+		# --set="image.repository=docker.io/$(IMG_NS)/operator" \
+		# --set="image.tag=$(BASE_IMG_TAG)" \
+		# --set="lurcher.image.repository=docker.io/$(IMG_NS)/lurcher" \
+		# --set="lurcher.image.tag=$(BASE_IMG_TAG)" \
```
fuhrmeistery: See Hook Makefile

fuhrmeistery:
Because This PR became so bloated, I decided to split this PR into multiple PRs
- #550 
- #551

PR #520, opened 2021-06-24
Validate Findings in Unit Tests

### LINKED ISSUE #518: Validate Findings with JSON Schema

Findings that are generated by SCB should be automatically validated so devs can rely on certain attributes existing, having a specific types, matching a specific pattern etc..

### Additional context
A common format for specifying and validating JSON is JSON Schema. However there exist many different versions (Drafts) to choose from. 

As the SCB Project uses OpenAPI 3.0 to specify Resources in Kubernetes we would like the JSON Schema to be compatible to the [OpenAPI JSON Schema](https://swagger.io/docs/specification/data-models/keywords/), that is basically JSON Schema Draft 05 with some changes. However we could not find a JavaScript Library that can validate the OpenAPI JSON Schema or JSON Schema Draft 05. 

However there exists a [JavaScript Library](https://www.npmjs.com/package/@openapi-contrib/json-schema-to-openapi-schema) to convert JSON Schema Draft 04 to OpenAPI 3.0 and [one](https://github.com/openapi-contrib/openapi-schema-to-json-schema) that does it in reverse. Also the biggest JavaScript Library for Validating JSON [Ajv](https://github.com/ajv-validator/ajv) supports Draft 04.  

Therefore the SCB Findings Format should be specified in JSON Schema Draft 04.

- [x] Create JSON Schema
- [x] Validate in Unit Tests
- [ ] Validate in Integration Tests

Also there are some Issues with the results that are currently produced that will be tracked here:
- Findings don't always contain a location
- Findings Locations should ideally be a uri/url but are currently not. Example: "localhost"

JohannesZahn:
## Description

Validate Findings in Unit Tests. This is not meant to introduce a reasonable findings schema, it simply implements the code to be able to validate the findings. Currently the findings schema is purposly forgiving so all parser pass the test as they are. 

### Checklist

* [x] Introduce findings-schema
* [x] Validate all Parsers in unit tests
* [x] Add new field ip_tcp_stack_layer that should replace osi_layer in the future


parser-sdk/nodejs/parser-utils.js
```diff
@@ -0,0 +1,41 @@
...
+async function validate(jsonData) {
+  const jsonSchemaString = await readFile(
+    __dirname + "/findings-schema.json",
+    "utf8"
+  );
+  const jsonSchema = JSON.parse(jsonSchemaString);
+  const validater = ajv.compile(jsonSchema);
+  const valid = validater(jsonData);
+  if (!valid) {
+    throw new Error(JSON.stringify(validater.errors, null, 2));
```
fuhrmeistery, line 29:
```suggestion
  const validator = ajv.compile(jsonSchema);
  const valid = validator(jsonData);
  if (!valid) {
    throw new Error(JSON.stringify(validator.errors, null, 2));
```

parser-sdk/nodejs/parser-utils.js
```diff
@@ -0,0 +1,41 @@
...
+    return {
+      ...finding,
+      id: uuid(),
+      parsed_at: new Date().toISOString(),
+    };
+  });
+}
+
+// eslint-disable-next-line
+async function validate(jsonData) {
```
fuhrmeistery, line 20:
```suggestion
async function validateJsonSchema(jsonData) {
```

JohannesZahn: What about validateWithJsonSchema? validateJsonSchema sounds to me like I am validating a json schema but I'm validating Json Data with a JSON Schema.

parser-sdk/nodejs/parser-utils.js
JohannesZahn: If I want to commit this, do I have to manually add the Signed-Off tag?

fuhrmeistery: I'm not sure what happens in this case. I would just try to commit it without any manual changes and look what happens. If the Bot complains, you could always rebase that commit.

parser-sdk/nodejs/parser-utils.js
fuhrmeistery: Does `validateAgainstJsonSchema` sound better?

parser-sdk/nodejs/parser-utils.js
```diff
@@ -33,9 +33,9 @@ async function validate(jsonData) {
...
 async function addSampleIdsAndDatesAndValidate(jsonData) {
   const extendedData = addIdsAndDates(jsonData);
-  validate(extendedData);
+  validateAgainstJsonSchema(extendedData);
 }
 
 module.exports.addIdsAndDates = addIdsAndDates;
-module.exports.validate = validate;
+module.exports.validate = validateAgainstJsonSchema;
 module.exports.validate_parser = addSampleIdsAndDatesAndValidate;
```
J12934, line 41:
```suggestion
module.exports.validateParser = addSampleIdsAndDatesAndValidate;
```

JohannesZahn: Done

parser-sdk/nodejs/parser-utils.js
JohannesZahn: it does! I changed it to `validateAgainstJsonSchema `

JohannesZahn:
Comments on Reviewing this PR:
- In every parser test that generates findings (usually every parser test) the method validateParser must be called to validate that the generated findings compy with the findings format. This will usually look like this:
``` 
const findings = await parse(fileContent); 
await expect(validateParser(findings)).resolves.toBeUndefined();
```
- The findings schema is very generic on purpose so the current parsers all pass the unit tests. The schema will be made more strict in another PR
- The library ajv-draft-04 which has rather few downloads/stars is used because it's "parent": ajv does not support draft04 out of the box but we want to use draft04 to be compatible with this [json schema to open api converter]( https://www.npmjs.com/package/@openapi-contrib/json-schema-to-openapi-schema).

parser-sdk/nodejs/parser-utils.js
```diff
@@ -0,0 +1,60 @@
+const fs = require("fs");
+const Ajv = require("ajv-draft-04");
+const ajv = new Ajv();
+const addFormats = require("ajv-formats");
+addFormats(ajv);
+const util = require("util");
+var jsonpointer = require("jsonpointer");
```
J12934, line 7:
```suggestion
const jsonpointer = require("jsonpointer");
```

parser-sdk/nodejs/parser-utils.js
```diff
@@ -0,0 +1,60 @@
...
+    return validateAgainstJsonSchema(jsonData);
+  }
+}
+
+function generateErrorMessage(errors, jsonData) {
+  for (let i = 0; i < errors.length; i++) {
+    errors[i] = { ...errors[i] };
+    const invalidValue = jsonpointer.get(jsonData, errors[i].instancePath);
+    errors[i].invalidValue = invalidValue;
+  }
```
J12934, line 54: Would fit better into the current code style do do this in a functional way.

```suggestion
  errors = errors.map((error) => {
    return { 
      ...error,
      invalidValue: jsonpointer.get(jsonData, error.instancePath),
    };
  });
```

parser-sdk/nodejs/parser-utils.js
```diff
@@ -0,0 +1,60 @@
...
+  const jsonSchema = JSON.parse(jsonSchemaString);
+  const validator = ajv.compile(jsonSchema);
+  const valid = validator(jsonData);
+  if (!valid) {
+    const errorMessage = generateErrorMessage(validator.errors, jsonData);
+    throw new Error(errorMessage);
+  }
+}
+
+// eslint-disable-next-line
```
J12934, line 38: what eslint rule is this disabling?

parser-sdk/nodejs/parser-utils.js
```diff
@@ -0,0 +1,60 @@
...
+async function addSampleIdsAndDatesAndValidate(jsonData) {
+  // add sample IDs and Dates only if the jsonData Array is not empty
+  if (jsonData.length > 0) {
```
J12934, line 41: Is this required?
Should the `addIdsAndDates` function also work fine with arrays with zero entries?

parser-sdk/nodejs/parser-utils.js
J12934: Same thing for the `validateAgainstJsonSchema` function.
When disabling eslint rules I always try to disable only the specif rule, so other people can understand what is supposed to be disabled. Like with the ` // eslint-disable-next-line security/detect-non-literal-fs-filename` above.

J12934 (changes requested):

parser-sdk/nodejs/parser-utils.js
JohannesZahn: They are not needed anymore and I removed them

parser-sdk/nodejs/parser-utils.js
JohannesZahn: it is not needed anymore and I removed it.

J12934 (approved):

PR #523, opened 2021-06-26
Add a Readme and License Notice to all secureCodeBox related DockerHub Repos

### LINKED ISSUE #474: Add a Readme and license notice to all secureCodeBox related DockerHub Repos
### Where to find the issue

All secureCodeBox related DockerHub repositories didn't have a Readme description and licence notice so far or otherwise it's outdated :(

### Describe the issue

As a _DockerHub User_ I would like to find a Readme description for each secureCodeBox repository which describes the docker image with all informations needed, especially the licence notice.

### Suggested change

There is a useful GitHub action that can be used for updating the Readme for each Repo: https://github.com/peter-evans/dockerhub-description

I would suggest to add this action to our pipeline and combine it additionally with the already working helm-docs action. We can use the helm-docs features to render specialised dockerHub Readme files which then can be published by the GH action mentioned above.

rseedorff:
## Description

This PR if applied fixes some appVersions in the HelmChart chart readme files.

Additional to that it introduces an extension to the existing HelmChart-Docs GitHub Workflow to create additional Readme files that can be used for all secureCodeBox DockerHub repos and as description for ArtifactHub Repos. Since the different Readme files are slightly different but mostly based on the same content, the creation of the different templates is based on template files.

All template files and generated Readme files are located in a newly created docs folder within each chart directory.

New Template files used by helm-docs:

- ./.helm-docs-template.gotmpl
- Readme.template.md.gotmpl

Closes #474

### Checklist


.helm-docs/templates.gotmpl
```diff
@@ -0,0 +1,106 @@
...
+[![License](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](https://opensource.org/licenses/Apache-2.0)
+
+Code of secureCodeBox is licensed under the [Apache License 2.0][scb-license].
+{{- end }}
+
+{{- define "extra.dockerLicenseSection" -}}
+## License
+[![License](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](https://opensource.org/licenses/Apache-2.0)
+
+View [license information](https://github.com/secureCodeBox/secureCodeBox/blob/master/LICENSE) for the software contained in this image.
```
J12934, line 71: I think this sentence is misleading to be included in the scanner images as these consist 100% of "non" secureCodeBox code.

.github/workflows/helm-docs.yaml
```diff
@@ -23,7 +23,6 @@ jobs:
           passphrase: ${{ secrets.GPG_COMMITS_PASSPHRASE }}
           git-user-signingkey: true
           git-commit-gpgsign: true
-          commit_options: '--signoff'
```
J12934, line 26: Looks accidental

.github/workflows/helm-docs.yaml
```diff
@@ -32,15 +31,101 @@ jobs:
...
+          # Verify installation
           ./helm-docs --version
 
       - name: Generate Helm Docs
         run: |
-          ./helm-docs/helm-docs
+          # Generate README.md based on Chart.yaml and template
+          ./helm-docs/helm-docs --template-files=./.helm-docs/templates.gotmpl --template-files=.helm-docs.gotmpl --template-files=./.helm-docs/README.md.gotmpl
+
+      - name: Generate Demo-Apps Docs
```
J12934, line 42: Would love it if we also included a way to run these locally for anybody who doesn't want to rely on the helm-docs bot. E.g. include a helm-docs target in the make file which runs these steps.

rseedorff: Me too :-) but it's not merged by now...

.github/workflows/helm-docs.yaml
rseedorff: Nope, is not. I fixed the signing because the option was placed at the wrong place/ action.

.helm-docs/templates.gotmpl
rseedorff: Gods point

.github/workflows/helm-docs.yaml
J12934: Mh ok right, something for the future then

J12934 (approved):

PR #532, opened 2021-07-01
Environment Variables can now be set for Parsers

JohannesZahn:
## Description

Added empty environment variable array to the parse definitions. Also removed the ncrack secrect in the parse definition as it is not required anymore according to @J12934. 

This will be used for #520, so users can decide if they want the program to crash if  the Findings validation detects invalid findings. This also enables integration tests for the Findings Validation. A environment variable can simply be set to true to make the tests fail if the validation fails. 

### Checklist

* [x] Make env an empty array


JohannesZahn:
As @J12934  suggested we should discuss in the review if we want to handle feature flags for parsers like this before merging.

J12934 (approved):

PR #538, opened 2021-07-03
Inherit environment variables, volumes, and volume mounts from parent scan

EndPositive:
## Description
This PR ensures that child scans inherit environment variables, volumes, and volume mounts (hereafter named as 'spec') from their parent scan (499c077, 0f224fa). The spec from the applied cascading rule is merged with those from the parent scan (cascading rule spec takes precedence 3391d74). Finally, to ensure that the spec from the cascading rule is only applied to the currently matched scan (and not its children), the child scan purges the cascading rule spec from the parent scan when inheriting them (b7b1856).

This behavior is useful in cases where you want to set the `http_proxy` variable for cascaded scans or when you want to mount CA certificates into cascaded scans.

I also considered retrieving the spec from the root scan, but currently there's no way (to my knowledge) to accurately determine the root scan in the cascading scan hook).

### Checklist


J12934:
Hi, sorry for the late feedback, has been a busy week :)

I think this makes sense and this is a valid usecase, but I don't think this should be enabled, but should be opted in by a setting on the scan, just like the inherit Labels and inherit Annotations settings: https://github.com/secureCodeBox/secureCodeBox/blob/main/operator/apis/execution/v1/scan_types.go#L19

```yaml
kind: Scan
metadata:
  name: "nmap-scanme.nmap.org"
spec:
  scanType: "nmap"
  parameters:
    - scanme.nmap.org
  env:
    - name: TEST_ENV
      valueFrom:
        secretKeyRef:
          key: secret-name
          name: zap-customer-credentials
  cascades:
    inheritLabels: false
    inheritAnnotations: true
    inheritVolumes: true
    inheritEnv: true
    matchLabels:
      securecodebox.io/intensive: light
```

The `inherit*` settings should already be cascaded to the child scans, so this should work for this use case.

EndPositive:
Done! I used cascades.inheritVolumes for both the volumes and volumeMounts, is that okay? Or do you prefer them separate?

hooks/cascading-scans/hook.ts
```diff
@@ -39,18 +42,26 @@ async function getCascadingRules(scan: Scan): Promise<Array<CascadingRule>> {
...
   parentScan: Scan,
   findings: Array<Finding>,
-  cascadingRules: Array<CascadingRule>
+  cascadingRules: Array<CascadingRule>,
+  cascadedRule: CascadingRule
 ): Array<ExtendedScanSpec> {
   let cascadingScans: Array<ExtendedScanSpec> = [];
   const cascadingRuleChain = getScanChain(parentScan);
 
+  parentScan = purgeCascadedRuleFromScan(parentScan, cascadedRule);
```
J12934, line 63: This took me a minute to understand why this is needed.
I think this should at least have a comment explaining why this is required or be moved closer to the point where the env vars & volumes are merged to make it clear that this is done to avoid duplicates there.

hooks/cascading-scans/hook.test.js
```diff
@@ -859,3 +875,675 @@ test("should copy proper finding ID into annotations", () => {
...
+
+  const cascadedScans = getCascadingScans(
+    parentScan,
+    findings,
+    sslyzeCascadingRules
+  );
+
+  const cascadedScan = cascadedScans[0]
+
+  expect(cascadedScans).toMatchInlineSnapshot(`
```
J12934, line 916: I think we should skip these assertions on the `cascadedScans` in these new tests, as these are basically a intermediate result for the purpose of this test, as the actual relevant logic on really happens in `getCascadingScanDefinition`. I was confused for a while looking at this and wondering why there is only on env var in the `cascadedScans` var.

That in itself might be something we should refactor in the future, unifying `getCascadingScans` and `getCascadingScanDefinition` and run all assertions on the actual scan specs. But this doesn't have to happen in this PR.

hooks/cascading-scans/hook.ts
```diff
@@ -25,8 +27,9 @@ interface HandleArgs {
 export async function handle({ scan, getFindings }: HandleArgs) {
   const findings = await getFindings();
   const cascadingRules = await getCascadingRules(scan);
+  const cascadedRule = await getCascadedRule(scan);
```
J12934, line 30: This var should probably have a different name to make it clearer what the distinction to the rules in `cascadingRules` is. Maybe something explicit like `cascadedRuleUsedForCurrentScan`.

J12934 (changes requested):
Hi this already looks great 👍
Found some small issues while going over the changes and some things which confused me.

hooks/cascading-scans/hook.test.js
EndPositive: I've updated the getCascadingScans function to return the actual scan definition. Also the tests have been updated, so please take another look.

hooks/cascading-scans/hook.ts
EndPositive: I've added some comments. Feel free to add more or move it where you think it fits better.

J12934 (approved):
Awesome work 🚀

PR #551, opened 2021-07-20
Add Makefile For Amass

fuhrmeistery:
- Add Makefile For Amass
- Move Amass Integration Tests Into Amass
- Exclude `*.tar` Files
- Add Image Build And Helm Deployment To Makefile
- Change Amass Build To Use Makefile
- See:
  -  #494 
  -  #476


J12934 (approved):

J12934:
Awesome 🚀

I think we should look into solutions to not have these Makefiles duplicated for every scanner once we start migrating the other scanners to this structure.

Another thing which seems to be missing is that the test coverage of the unit test is not getting submitted to code climate, or did i miss that?

@rseedorff can you look into why fossa / license scanner is failing?
Or @fuhrmeistery do you have access to it? My fossa is still broken 😕

fuhrmeistery:
> I think we should look into solutions to not have these Makefiles duplicated for every scanner once we start migrating the other scanners to this structure.

I agree, but I did not find any leads yet. :+1: 

> Another thing which seems to be missing is that the test coverage of the unit test is not getting submitted to code climate, or did i miss that?

Good point. No, code coverage is currently not submitted to code climate for Makefile builds at the moment.
I think regarding our plans to only execute Build Jobs for Code that has changes, we might run into some problems regarding codeclimate. Their [documentation](https://docs.codeclimate.com/docs/configuring-test-coverage#section-multiple-test-suites) for configuring test coverage states that for parallel/multiple test/test-suits all test-suites need to be aggregated that codeclimate can combine them. Because codeclimate requires branch name and commit hash, we could run into trouble because we won't have coverage for everything in every commit. 

> @rseedorff can you look into why fossa / license scanner is failing?
> Or @fuhrmeistery do you have access to it? My fossa is still broken confused
Are these issues resolved? At least, I don't see any issues with the checks on this PR.

PR #556, opened 2021-07-21
Fix inconsistent casing in findings created in DefectDojo Hooks `syncFindingsBack` Mode

J12934:
Severities were accidentally formatted incorrectly by using the jackson annotations wrongly.

### DIFF
--- a/hooks/persistence-defectdojo/hook/src/main/java/io/securecodebox/persistence/models/SecureCodeBoxFinding.java
+++ b/hooks/persistence-defectdojo/hook/src/main/java/io/securecodebox/persistence/models/SecureCodeBoxFinding.java
@@ -42,14 +42,10 @@ public class SecureCodeBoxFinding {
   Map<String, Object> attributes;
 
   public enum Severities {
-    @JsonProperty("HIGH")
-    High,
-    @JsonProperty("MEDIUM")
-    Medium,
-    @JsonProperty("LOW")
-    Low,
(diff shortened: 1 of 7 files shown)


hooks/persistence-defectdojo/hook/src/main/java/io/securecodebox/persistence/models/SecureCodeBoxFinding.java
```diff
@@ -42,14 +42,10 @@
...
-    @JsonProperty("MEDIUM")
-    Medium,
-    @JsonProperty("LOW")
-    Low,
-    @JsonProperty("INFORMATIONAL")
-    Informational
+    HIGH,
+    MEDIUM,
+    LOW,
+    INFORMATIONAL
```
fuhrmeistery, line 48: You mentioned that the Jackson annotations were used wrongly.
Was it intentional to remove these annotations all together?

J12934: I'm not sure what causes this to behave wrongly, as this seems to work when deserializing JSON Findings to java objects but seems to fail during serialization. The docs didn't indicate to me that this would be a problem.

Naming the enums exactly like their actual value still seems like the easiest solution, and is even more consistent with the usual naming schema of java enums.

fuhrmeistery (approved):

PR #583, opened 2021-08-09
Cascading Scans bugfixing

EndPositive:
## Description
- 0471226 fixes a bug where if a single rule would be applied to two findings, it would only be templated using the first finding.
- 545898e simple cleanup.
- 5896910 fixes a bug where the new secureCodeBox.io annotations (such as parent-scan, matched-finding, and chain) would be overwritten by the parent's annotations (if inheritAnnotations was enabled; default).

### Checklist


J12934 (approved):
Oh that's a nasty bug, good catch 👍

PR #592, opened 2021-08-11
Integrating new scanner Whatweb to the secureCodeBox (closes #567)

### LINKED ISSUE #567: Add WhatWeb as new scanner for fingerprinting usecases
### Is your feature request related to a problem

As security analyst i would like to the secureCodeBox for testing my own external attack surface. Therefore it would be helpful to add a scanner for fingerprinting HTTP Service in a more detailed way. The WhatWeb Scanners seems to be a good first candidate for that.

Integrate WhatWeb as new SCB scanner with cascadingScan rules matching HTTP services (AMASS -> NMAP -> WhatWeb). 

### Additional context

* https://www.whatweb.net/
* https://github.com/urbanadventurer/WhatWeb

## Steps to implement a new scanner
Hint: A general guide how to implement a new SCB scanner is documented [here](https://docs.securecodebox.io/docs/contributing/integrating-a-scanner)

- [x] Create a new folder with the name of the [scanner here](https://github.com/secureCodeBox/secureCodeBox/tree/master/scanners)
- [x] Add a `README.gotmpl` and give a [brief overview](https://docs.securecodebox.io/docs/contributing/integrating-a-scanner/readme) of the scanner and its configuration options.
- [x] Add a HelmChart and document all configuration options.
- [x] Implement a new scanner specific `scan-type.yaml`
- [x] Implement a new scanner specific `parse-definition.yaml`
- [x] Add (optional) some `cascading-rules.yaml` like documented [here](https://docs.securecodebox.io/docs/api/crds/cascading-rule)
- [x] Add (optional) a `Dockerfile` for the scanner if there is no existing one publicly available on dockerHub
- [x] Use the [parser-SDK](https://github.com/secureCodeBox/secureCodeBox/tree/master/parser-sdk) to implement a new findings parser (currently based on NodeJS)
- [x] Add unit tests with at minimum 80% test coverage
- [x] Add some example `scan.yaml` and `finding.yaml` files in the [example folder](https://docs.securecodebox.io/docs/contributing/integrating-a-scanner/examples-dir)
- [x] Implement a [new integration or E2E test](https://docs.securecodebox.io/docs/contributing/integrating-a-scanner/integration-tests) for the hook [here](https://github.com/secureCodeBox/secureCodeBox/tree/master/tests/integration)

SebieF:
## Description

This PR adds the new fingerprinting scanner [Whatweb](https://morningstarsecurity.com/research/whatweb) to the scb. It identifies websites and recognizes used technology. This can be especially useful for cascading scans.

### Checklist

(closes #567)


J12934:
Is there a reason to use the whatweb xml over the json output? @SebieF 
XML is always a bit harder as it requries extra libs in nodejs.
Without the xml the parser could probably live completly without external dependencies.

scanners/whatweb/parser/parser.js
```diff
@@ -0,0 +1,99 @@
...
+  const hosts = await parseResultFile(fileContent);
+  return transformToFindings(hosts);
+}
+
+function transformToFindings(targets) {
+
+  const targetFindings = targets.map(target => {
+    let tempFinding = {
+      name: target.uri,
+      category: "URL", //TODO?
```
J12934, line 17: Maybe "Web Application"?
As whatweb is meant to identify / analyse web apps?

scanners/whatweb/parser/parser.js
```diff
@@ -0,0 +1,99 @@
...
+          if(target.plugin) {
+            target.plugin.map(plugin => {
+              if (plugin.name[0] === "IP")
+              newTarget.ipAddress = plugin.string[0];
+              else if (plugin.name[0] === "Title")
+              newTarget.title = plugin.string[0];
+              else
+              newTarget.additional.push(plugin)
+            });
+          }
```
J12934, line 88: Would recommend to only use Array.map when acutally mapping from one array to another.
This uses it to just loop over the array.

```suggestion
          if(target.plugin) {
            for (const plugin of target.plugin) {
              if (plugin.name[0] === "IP")
              newTarget.ipAddress = plugin.string[0];
              else if (plugin.name[0] === "Title")
              newTarget.title = plugin.string[0];
              else
              newTarget.additional.push(plugin)
            }
          }
```

scanners/whatweb/parser/parser.js
```diff
@@ -0,0 +1,99 @@
...
+}
+
+function transformToFindings(targets) {
+
+  const targetFindings = targets.map(target => {
+    let tempFinding = {
+      name: target.uri,
+      category: "URL", //TODO?
+      description: target.title,
+      location: target.ipAddress,
```
J12934, line 19: Location needs to be a full url in the scb v3 finding model.
Does using the url from the target work? Or are there case where it would not be defined?

```suggestion
      location: target.uri,
```

J12934 (changes requested):

SebieF:
> Is there a reason to use the whatweb xml over the json output? @SebieF
> XML is always a bit harder as it requries extra libs in nodejs.
> Without the xml the parser could probably live completly without external dependencies.

Yes, I checked the json output and it was pretty badly formatted imo, so I went for the xml log.

scanners/whatweb/parser/parser.js
SebieF: I replaced target.ipAddress with target.uri for the location parameter to match the v3 finding model. This results in the situation that both name and uri state the domain in the finding, and ipAddress is moved to attributes. But I think it's best we can do?

SebieF:
> > Is there a reason to use the whatweb xml over the json output? @SebieF
> > XML is always a bit harder as it requries extra libs in nodejs.
> > Without the xml the parser could probably live completly without external dependencies.
> 
> Yes, I checked the json output and it was pretty badly formatted imo, so I went for the xml log.

However, json output is also fine, so I will reduce the overhead in the parser by xml.

scanners/whatweb/scanner/Dockerfile
```diff
@@ -0,0 +1,13 @@
+# SPDX-FileCopyrightText: 2021 iteratec GmbH
+#
+# SPDX-License-Identifier: Apache-2.0
+
+FROM debian:latest
+
+RUN apt-get update \
+    && apt-get install whatweb -y \
```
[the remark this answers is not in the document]
SebieF: I added the version tag and changed the Docker Container so that it now pulls from ruby and clones the git repository with the latest version tag. Unfortunately, there is no "latest" tag, so at the moment we have to manually update it..

rseedorff (dismissed):

rseedorff:
Hi @SebieF, 
Great work 👍
One thing last: your PR has a conflict that needs to be solved... than the PR should be ready to be merged

.github/workflows/ci.yaml
```diff
@@ -1052,6 +1054,17 @@ jobs:
...
           npx jest --ci --color scanner/sslyze.test.js
 
+      # ---- Whatweb Integration Tests ----
+
+      - name: "Whatweb Integration Tests"
+        run: |
+          kubectl -n integration-tests delete scans --all
+          helm -n integration-tests install whatweb ./scanners/whatweb/  \
+            --set="parser.image.tag=sha-$(git rev-parse --short HEAD)" \
+            --set="parser.image.repository=docker.io/${{ env.DOCKER_NAMESPACE }}/parser-whatweb"
```
rseedorff, line 1064:
```suggestion
          helm -n integration-tests install whatweb ./scanners/whatweb/  \
            --set="parser.image.tag=sha-$(git rev-parse --short HEAD)" \
            --set="parser.image.repository=docker.io/${{ env.DOCKER_NAMESPACE }}/parser-whatweb"
            --set="parser.env[0].name=CRASH_ON_FAILED_VALIDATION" \
            --set-string="parser.env[0].value=true"
```

rseedorff (changes requested):

PR #594, opened 2021-08-12
Update NPM Dependencies

twwd:
Updates NPM dependencies to the latest compatible versions including
version 6.1.8 of `tar` that contains security fixes.

In addition, fixes that npm 6 was used instead of npm 7 in unit tests and an outdated command to transpile TypeScript.

### Checklist

### DIFF
--- a/hooks/finding-post-processing/hook/package.json
+++ b/hooks/finding-post-processing/hook/package.json
@@ -4,7 +4,6 @@
   "description": "secureCodeBox Finding Post Processing Hook",
   "main": "hook.js",
   "scripts": {
-    "build": "npx typescript hook.ts --sourceMap",
     "test": "jest ."
   },
   "repository": {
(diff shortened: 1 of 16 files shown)


hooks/finding-post-processing/hook/package.json
```diff
@@ -4,7 +4,6 @@
   "description": "secureCodeBox Finding Post Processing Hook",
   "main": "hook.js",
   "scripts": {
-    "build": "npx typescript hook.ts --sourceMap",
```
J12934, line 7: Why is the script removed here, but updated in the cascading-scans hook?
@twwd

twwd: The `find-post-processing` hook is written in plain JavaScript and thus requires no TypeScript transpilation. In turn, the `cascading-scans` hook is written in TypeScript like the `notification` hook (which was already updated to `tsc`).

J12934: Haha ok good point.

Lost track what typescript and what is not :)

J12934 (approved):

PR #596, opened 2021-08-12
DefectDojo Hook Tests now pass on any OS

JohannesZahn:
## Description

Unit Tests on Windows failed even though they were supposed to pass. This was due asserting JSON Objects as equal that contained different line breaks.  This was fixed by using the Library JSONAssert.

### Checklist


J12934:
Could we remove the Json assert Lib by parsing the Json into our Finding Class and running the assertions on these?

JohannesZahn:
It  should be possible unless Jackson changes the order of the list which I am not completly sure on right now. I am also not sure  if we should then use Jacksons ObjectNode.equals() on the JSON Findings or rely on a custom/lombok generated Equals methods on the Objects themselves. The second approach would look somewhat like this

```java
DefectDojoImportFinding[] expectedDDFindings = jsonMapper.readValue(ddFindingsUrl, DefectDojoImportFinding[].class);
DefectDojoImportFinding[] actualDdFindings = jsonMapper.readValue(actualDdFindingsString, DefectDojoImportFinding[].class);
assertArrayEquals(expectedDDFindings,actualDdFindings);
```

JohannesZahn:
This moved to #600

PR #597, opened 2021-08-12
Revert Change in Helmignore to Include the Scripts Again

J12934:
This caused the configmaps to contain no scripts, as they were ignored from the charts :(

### DIFF
--- a/scanners/zap-advanced/.helmignore
+++ b/scanners/zap-advanced/.helmignore
@@ -11,7 +11,8 @@ scanner/zapclient/
 scanner/tests/
 scanner/venv/
 scanner/.pytest_cache/
-scanner/.idea/examples/
+scanner/.idea/
+examples/
 docs/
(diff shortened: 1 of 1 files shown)


scanners/zap-advanced/.helmignore
```diff
@@ -5,8 +5,13 @@
 .DS_Store
 
 parser/
-scanner/
-examples/
```
rseedorff, line 9: Examples should be excluded...

rseedorff (changes requested):

scanners/zap-advanced/.helmignore
J12934: Ah, forgot a new line, the `examples/` exclude was accidentally put into the same line as the `.idea/` making both worthless. 👍

rseedorff (approved):

PR #600, opened 2021-08-18
Improved Test Structure and Tests in DefectDojo Hook

JohannesZahn:
## Description

The Mapper converting SecureCodeBox to DefectDojo Findings accepted and returned JSON Strings and could therefore not easily be tested. Now the Mapper only Converts secureCodeBox Findings Objects to defectDojo Findings Objects. There were previously some tests for this mapping but they were enhanced by seperating tests for mapping dates, descriptions and basic fields.

A Test Class for the ScanResultService (which returns the ScanResults that are passed to DefectDojo) was introduced that is now resposible for testing that the assembled scan results that will be uploaded to defectDojo are  processed correctly.

Furthermore the PersistenceConfig now features a field to specify a Timezone for parsing dates instead of using the default timezone on a machine. Therefore the timezone can be set in the unit tests to ensure they produce the same results regardles of the location

### Checklist


JohannesZahn:
@J12934 this is ready for review, however one check does not complete, I tried re-running all jobs but it didn't seem to help.

hooks/persistence-defectdojo/hook/src/main/java/io/securecodebox/persistence/service/scanresult/GenericParserScanResultService.java
```diff
@@ -0,0 +1,54 @@
...
+   * the DefectDojo Generic JSON Parser. This result as a string is then returned together with a filename
+   * in a ScanFile object. The ending of the filename is essential as it is evaluated by DefectDojo
+   * @param ppConfig config where the location of the scan result is specified
+   * @return
+   * @throws IOException
+   * @throws InterruptedException
+   */
+  @Override
+  public ScanFile getScanResult(PersistenceProviderConfig ppConfig) throws IOException, InterruptedException {
+    LOG.debug("No explicit Parser specified using Findings JSON Scan Result");
```
J12934, line 43:
```suggestion
    LOG.debug("No explicit Parser specified. Using Findings JSON Scan Result");
```

hooks/persistence-defectdojo/hook/src/main/java/io/securecodebox/persistence/service/scanresult/SpecificParserScanResultService.java
```diff
@@ -0,0 +1,36 @@
...
+   * in a ScanFile object. The ending of the filename is essential as it is evaluated by DefectDojo.
+   * This is usually used when DefectDojo natively supports the scanner that produced the raw result.
+   * @param ppConfig config where the location of the scan result is specified
+   * @return
+   * @throws IOException
+   * @throws InterruptedException
+   */
+  @Override
+  public ScanFile getScanResult(PersistenceProviderConfig ppConfig) throws IOException, InterruptedException {
+    LOG.debug("Explicit Parser is specified, using Raw Scan Result");
```
J12934, line 31:
```suggestion
    LOG.debug("Explicit Parser is specified. Using Raw Scan Result");
```

J12934 (approved):
Looks good 👍

Added some minor suggestions to log lines which might be a bit confusing without punctuation.

JohannesZahn:
I commited your suggestions @J12934 :)

J12934 (approved):

PR #601, opened 2021-08-18
Introduce Makefile For All Scanners

EndPositive:
## Description
This PR continues @fuhrmeistery 's work on the scanner Makefiles.

I have moved the Makefile for Amass into the project's root directory. From each scanner, we include the root Makefile and override configuration variables (such as scanner name) or whether a scanner container needs to be built. Each scanner can specify what demo target dependencies it has.

https://github.com/EndPositive/secureCodeBox/blob/eec9b75edb70b347af1e47dcdbc7cb7e14562e2b/scanners/zap/Makefile#L8-L13

Finally, in this manner, we can also completely override the root makefile's behaviour. Zap-advanced's Makefile, for example, redirects the parser-related targets to zap's makefile. Kubeaudit overrides the helm deploy target such that it can set extra helm values.

### Checklist
* [x] Reduce the amount of duplicated code in the makefiles by using 'some sort of inheritence'
* [x] Adapt Makefile to add make targets for building custom/third party scanners
* [ ] Add unit tests for scanners (detect java/python)
* [x] Replace CI amass job by Github Matrix / run them one by one in a single job (open for discussion; I believe kind cluster create is pretty expensive in the pipeline and a matrix job might explode that time..; on the other hand matrix would reduce oveall pipeline duration if the pipeline has enough resources)


EndPositive:
Extra note of caution: the latest commit ec8c8f694e63271cb01e2aab458f1e9ec30b0a42 updates the CI so that all scanners are built with the new Makefiles. It would be really nice if someone could give it a review and trigger a run.

J12934:
> Replace CI amass job by Github Matrix / run them one by one in a single job (open for discussion; I believe kind cluster create is pretty expensive in the pipeline and a matrix job might explode that time..; on the other hand matrix would reduce oveall pipeline duration if the pipeline has enough resources)

Yes good point. We've planned to change the pipeline so that it looks for changed files in the relevant folder. So that we would only run the builds for a specific scanner if files for it were changed, or if a central component e.g. operator, lurker, parser-sdk were changed. This should speed this up quite a lot as most pipelines would only run the build and tests for the changed scanner(s).

> I remember that in the previous set-up, the integration tests for some scanners were disabled. Do we want to keep that behavior?

I think the only intentionally disabled integration-tests are the ones for the zap-advanced as they were just taking too long to run the scans. That should hopefully also be kinda solved by only running the tests only when the acutal zap-advanced code was changed.

> In the ZAP advanced makefile, the ZAP parser is automatically built in the docker-build target. In the pipeline, we instead use the docker build push action and detect whether to build the parser/scanner based on the presence of their Dockerfile. Any good ideas to ensure that in the pipeline the ZAP parser is built before the ZAP advanced job starts?

Ah yes, thats a kind of edge case 🤔 Could we overwrite the build parser make targets in the zap-advanced makefile enought to just build the parser form the zap folder? I think this would be easier than reusing the image build for zap.  

Checked the pr out and tried the nmap makefile looks great already 🥳
I think there needs to be a small fix to the parser js test imports, will push these and trigger the ci.

.github/workflows/ci.yaml
```diff
@@ -360,9 +360,9 @@ jobs:
 
       # Decide which Docker Meta job to run based on whether this scanner is a third party scanner, whose tag should be the Chart's appVersion
       - name: Docker Meta
-        id: docker_meta
+        id: docker_meta_scanner
+        if: hashFiles(format('./scanners/{0}/scanner/Dockerfile', matrix.unit)) != ''
         uses: crazy-max/ghaction-docker-meta@v1
-        if: contains(fromJson('["angularjs-csti-scanner", "gitleaks", "kube-hunter", "kubeaudit", "ncrack", "nmap"]'), matrix.unit)
```
EndPositive: I believe this manual check was necessary. For example, the git-repo-scanner has its own Dockerfile but is always tagged with the secureCodeBox tag, and not the appVersion (which is 1.0 for git-repo-scanner).

In the old Pipeline, there was a distinction between third-party scanners and custom scanners. Both have a Dockerfile, but third-party scanners follow their own versioning, and custom scanners follow secureCodeBox versioning.

EndPositive:
> Yes good point. We've planned to change the pipeline so that it looks for changed files in the relevant folder. So that we would only run the builds for a specific scanner if files for it were changed, or if a central component e.g. operator, lurker, parser-sdk were changed. This should speed this up quite a lot as most pipelines would only run the build and tests for the changed scanner(s). 

That would be awesome. For now, lets keep this PR with a matrix job and once we're fully over to Makefiles, we can do the extra magic.

> I think the only intentionally disabled integration-tests are the ones for the zap-advanced as they were just taking too long to run the scans. That should hopefully also be kinda solved by only running the tests only when the acutal zap-advanced code was changed.

You're right that zap-advanced was manually disabled, however I noticed that git-repo-scanner was never included. I can imagine that was due to GitHub rate-limiting?

> Ah yes, thats a kind of edge case thinking Could we overwrite the build parser make targets in the zap-advanced makefile enought to just build the parser form the zap folder? I think this would be easier than reusing the image build for zap.

Yep, that's currently what happens in the Makefile for zap-advanced, check it out! However, in the pipeline, we make use of the docker action, skipping the Makefile's build target altogether. For your proposal to work in the pipeline, we would have to get rid of the Docker build action. Do you know if the ubuntu-latest VM runs `docker build` commands properly? In any case, that would also require the Makefile to also manage the tags and pushes right?

> Checked the pr out and tried the nmap makefile looks great already partying_face
> I think there needs to be a small fix to the parser js test imports, will push these and trigger the ci.
Great!

J12934:
> You're right that zap-advanced was manually disabled, however I noticed that git-repo-scanner was never included. I can imagine that was due to GitHub rate-limiting?

yeah rate limiting is a problem, also we try to keep the integration test completly offline, that pretty hard to do with the git-repo-scanner.  We should "just" be able to disable these tests for now by overwriting the `integration-tests` make target in the makefiles for these scanners to a empty script, right? As a temporary workaround, running these only when the code was changed should hopefully decrease the number of requests enough to be pretty stable.

> Yep, that's currently what happens in the Makefile for zap-advanced, check it out! However, in the pipeline, we make use of the docker action, skipping the Makefile's build target altogether. For your proposal to work in the pipeline, we would have to get rid of the Docker build action. Do you know if the ubuntu-latest VM runs docker build commands properly? In any case, that would also require the Makefile to also manage the tags and pushes right?

Mhh yeah... I'm not sure how easily we can / should migrate from the action, i does a lot of neat stuff in the background. I'll try to see if I can came up with a good solution to this 🤔

EndPositive:
Awesome work @J12934 !

Do you guys want to make a release (with the new scanners) before we start messing up main and the pipeline?

Also, let's try to keep the work a bit separated, so contributing becomes easier. Keep this PR only on creating makefiles for scanners (no pipeline changes) and then I'll create another PR for hook makefiles. Maybe you can then look at updating the pipeline, since I'm unable to properly test not being authorized to trigger pipelines.

EndPositive:
Since this PR has gotten too bloated with further pipeline testing, I've decided to completely separate the changes for Makefiles and those for the pipeline. I'll close this PR in favor of #622 and further PR's.

PR #603, opened 2021-08-20
Finding are validated in Integration Tests

JohannesZahn:
## Description

The recently introduced Finding Validation should also be applied in the integration tests.
This can simply be implemented by setting an environment variable which is evaluated in the parser sdk and makes the scanner crash if the findings are invalid.

### Checklist


Weltraumschaf (approved):
LGTM

PR #605, opened 2021-08-20
Added new security scanner: nuclei

rseedorff:
## Description

This PR if applied integrates a new security scanner with the secureCodeBox: nuclei

* https://nuclei.projectdiscovery.io/
* https://github.com/projectdiscovery/nuclei

### Checklist


scanners/nuclei/parser/parser.js
```diff
@@ -0,0 +1,60 @@
...
+
+function getAdjustedSeverity(severity) {
+  return severity === "CRITICAL"
+    ? "HIGH"
+    : severity === "INFO"
+    ? "INFORMATIONAL"
+    : severity === "UNKNOWN"
+    ? "INFORMATIONAL"
+    : severity;
+}
```
SebieF, line 46: Adjust function to make codeclimate happy
```suggestion
function getAdjustedSeverity(severity) {
    if(severity === "CRITICAL") 
         return "HIGH";
    if(severity === "INFO" || severity === "UNKNOWN")
         return "INFORMATIONAL";
    return severity;     
}
```

scanners/nuclei/parser/parser.test.js
```diff
@@ -0,0 +1,70 @@
...
+  const fileContent = 
+    await readFile(
+      __dirname + "/__testFiles__/empty-test.jsonl",
+      {
+        encoding: "utf8",
+      }
+    )
+  
+  const findings = await parse(fileContent);
+  //await expect(validateParser(findings)).resolves.toBeUndefined();
```
SebieF, line 26: Why is validating the results commented out?

SebieF (changes requested):
Nicely done :)

SebieF (approved):

SebieF:
When I try to install the scanner via helm, I get the following error:
`helm upgrade --installnuclei scanners/nuclei/`

`Release "nuclei" does not exist. Installing it now.
Error: ScanType.execution.securecodebox.io "nuclei" is invalid: spec.jobTemplate.spec.activeDeadlineSeconds: Invalid value: "null": spec.jobTemplate.spec.activeDeadlineSeconds in body must be of type integer: "null"
`

Seems like the value for activeDeadlineSeconds must be set manually before the scanner can be installed. I guess this behavior is not intended?

scanners/nuclei/parser/parser.js
```diff
@@ -0,0 +1,58 @@
...
+  });
+}
+
+function getAdjustedSeverity(severity) {
+  return severity === "CRITICAL"
+    ? "HIGH"
+    : severity === "INFO" || severity === "UNKNOWN"
+    ? "INFORMATIONAL"
+    : severity;
+}
```
J12934, line 44: Can we change that to if / switch syntax?
Double ternary statements are really hard to read 😕

J12934:
Wouldn't the cache volume currently block multiple nuclei scanners from running at the same time?
To my understanding ReadOnlyMany can only be mounted multiple times when the volume mount is marked as readOnly which doesn't seem to be the case.

And should the cache volume be filled by something like a CronJob which periodically refreshes the cache and the scans having this volume mounted as readOnly?

Also should this be turned on by default, as many storage k8s environments don't support ReadOnlyMany: https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes

rseedorff:
@J12934 Yes you are right, in practice this happens sometimes, the nuclei scan then will be executed in a more serialized way, but is is not always the case 🤷🏻‍♂️

I already tried to find another solution at the nuclei project side, but this will take some time to finish: https://github.com/projectdiscovery/nuclei/issues/895#

PR #616, opened 2021-08-30
Bugfixed scanner parseDefinitions ttlSecondsAfterFinished HelmValue inconsistencies 

SebieF:
## Description

We had a lot of new scanner pull requests recently, that's why I came to think about code/file consistency for our scanners. I think it is worth discussing this topic publicly or in our retro, because it makes our code more maintainable and easier to understand. 
For now, I only fixed some small inconsistencies and a typo for templates/scanner-name-parse-definition.yaml.

### DIFF
--- a/scanners/amass/templates/amass-parse-definition.yaml
+++ b/scanners/amass/templates/amass-parse-definition.yaml
@@ -8,6 +8,6 @@ metadata:
   name: "amass-jsonl"
 spec:
   image: "{{ .Values.parser.image.repository }}:{{ .Values.parser.image.tag | default .Chart.Version }}"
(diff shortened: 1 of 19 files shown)


J12934:
+1 for consistency.

But placing the ttl after finished would be conceptually wrong for me, as the ttl has nothing to do with the image, but the kubernetes job created by the parseDefinitions. I think it is fine where it it.

SebieF:
> +1 for consistency.
> 
> But placing the ttl after finished would be conceptually wrong for me, as the ttl has nothing to do with the image, but the kubernetes job created by the parseDefinitions. I think it is fine where it it.

Fair point, wasn't sure about that. In hindsight, I think only ncrack used this .image. so it makes more sense to adjust that.

rseedorff (approved):

Issue #610, opened 2021-08-27
🏗 Migrate Nikto Build to be in line with other 3rd party scanners


J12934:
We maintain a number of docker images / dockerfiles for third party tools, most of these work pretty similarly by consisting of a singular dockerfile which can be easily build by using docker build.

Nikto is currently the only outlyer requiring a more complex setup and a locally cloned nikto repository.

This should be changed so that the nikto images can be build like the other custom tool images, so that it doesn't need a special pipeline setup but can be build in a matrix build setup.

PR #622, opened 2021-09-01
Introduce Makefile For All Scanners & Hooks

EndPositive:
## Description
This PR continues the effort of #601, however only applies changes unrelated to the pipeline. This makes intermediate merging, reviewing and testing easier.

This PR does NOT break the current pipeline and works as is locally.

This PR contains the following changes

* [x] Created generic scanner makefile
* [x] Implemented makefiles for all scanners (except Nikto, but won't implement until #610 is solved)
* [x] Implemented makefiles for all hooks
* [x] Added unit-test targets for Java
* [x] Added unit-test targets for Python
* [x] Update cleanup target to also clean cluster
* [x] Add target to existing root Makefile to run all tests

### Checklist


J12934:
Awesome work, yeah good idea to keep the Makefile and pipeline changes seperate for now to not block releases.
Can I / we help you in any way to complete the remaining tasks in this PR?

EndPositive:
I'm planning to finish this during the weekend and I'll tag you for a review on/before Monday!

Although not necessary for this PR, it would be really nice if you could take a look at #610.

EndPositive:
@J12934 I can't see any more Makefiles.. I think that should do it, though. Could you take a look and make changes wherever neccessary?

J12934:
Yes 👍
Will do, thank you for the awesome work on this 👌

tests/integration/helpers.js
```diff
@@ -173,8 +173,6 @@ async function scan(name, scanType, parameters = [], timeout = 180) {
  * @returns {scan.findings} returns findings { categories, severities, count }
  */
 async function cascadingScan(name, scanType, parameters = [], { nameCascade, matchLabels }, timeout = 180) {
-  namespace = "cascading-tests";
-
```
EndPositive, line 177: Pipeline is failing because I changed the namespace for cascading scans here. I did that to keep the Makefile for cascading scans easy, but I'll take a look again.

J12934: Already looking into it :)
Changed it to run in the integration-tests namespace in the pipeline.
"Should be working now"™️

Currently only on this branch, will push to the pr when the pipeline passes: https://github.com/secureCodeBox/secureCodeBox/commits/makefile-scanners

J12934: Seems to be working now.
Anything else from you point of view stopping this from getting merged?
Otherwise i think we can merge this asap

EndPositive:
@J12934 awesome, thank you for the fixes! 👏

The Makefile for Nikto is still missing, but we can fix that once it's version is pinned.

Another thing I wanted to mention is the integration tests for cascading scans. Currently they are set up in a manner that requires them to run separately (due to different scanner installs; i.e. ncrack vs sslyze). If you look at the Makefile for cascading scans, you'll notice there's also a test-2 target. With this setup, that test-2 won't run in a matrix job. Did you have any thoughts on that?

Besides that, I think it's ready to be merged. It's bound to have issues, but let's give it a shot. It's only for development purposes currently anyways.

I'll take a look at updating the integration tests documentation coming week.

J12934 (approved):

PR #632, opened 2021-09-08
Added a Typo3 demo-target

Ilyesbdlala:
## Description

This PR adds a Typo3 demo-target. A Dockerfile and a corresponding Helm chart are included. 
This allows running cascading scans and integration tests with the Typo3scan scanner.
The demo-target Image has to be uploaded to the secureCodeBox docker hub before merging this to the main branch.

### Checklist

* [X] Test your changes as thoroughly as possible before you commit them. Preferably, automate your test by unit/integration tests.
* [X] Make sure `npm test` runs for the whole project.
* [X] Make codeclimate checks happy


demo-targets/old-typo3/Chart.yaml
```diff
@@ -0,0 +1,20 @@
+# SPDX-FileCopyrightText: 2021 iteratec GmbH
+#
+# SPDX-License-Identifier: Apache-2.0
+
+apiVersion: v2
+version: v3.1.0-alpha1
+appVersion: "v1.0"
```
J12934, line 7: Maybe we could make this match the typo3 version used (v9.5.26).
This is used as the default image tag and it would be nice to have this match the actual typo3 version of the chart & image.

```suggestion
appVersion: "v9.5.26"
```

Ilyesbdlala: Yeah that is better. I'll commit it in

J12934 (approved):
Build and pushed the image to docker.io/securecodebox/old-typo3:v9.5.26:

https://hub.docker.com/repository/docker/securecodebox/old-typo3

PR #635, opened 2021-09-10
Upgrade all npm packages with npm audit fix

J12934:
Upgraded all npm dependencies, with our `./bin/npm-audit-fix-all.sh` script.
This should replace the open snyk prs.

Closes #626
Closes #627
Closes #628
Closes #629
Closes #631

### DIFF
--- a/hook-sdk/nodejs/package-lock.json
+++ b/hook-sdk/nodejs/package-lock.json
@@ -4456,9 +4456,9 @@
       "dev": true
     },
     "node_modules/tar": {
-      "version": "6.1.8",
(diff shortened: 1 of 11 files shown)


SebieF (approved):
It seems to be a bit odd that the package.json are now oneliners, but could also be the display on GitHub?

J12934:
Mh that't weird, I'll take a look on why that is happening...

J12934:
Ok should be correct now, seems like npm was only doing it for package-json where it found the formatting to be weird.

Still kind weird behaviour of npm...

SebieF (approved):

PR #642, opened 2021-09-13
Use Multi-Stage-Build For Nikto

### LINKED ISSUE #610: 🏗 Migrate Nikto Build to be in line with other 3rd party scanners

fuhrmeistery:
If applied, this PR will:
- change the Dockerfile for nikto to a multi-stage Dockerfile
- remove custom nikto build job
- add nikto to the build matrix for third-party scanners
- closes #610

Signed-off-by: Yannik Fuhrmeister <yannik.fuhrmeister@iteratec.com>

### DIFF
--- a/scanners/nikto/scanner/Dockerfile
+++ b/scanners/nikto/scanner/Dockerfile
@@ -16,11 +16,11 @@ COPY wrapper.sh /wrapper.sh
 RUN apk add --update --no-cache --virtual .build-deps \
      perl \
      perl-net-ssleay \
-  && addgroup nikto \
-  && adduser -G nikto -s /bin/sh -D nikto
+  && addgroup -g 1001 nikto \
+  && adduser -G nikto -s /bin/sh -D -u 1001 nikto
 
 COPY --from=build --chown=nikto:nikto /nikto/program /nikto
 
-USER nikto
+USER 1001 
 
 ENTRYPOINT [ "sh", "/wrapper.sh" ]
--- a/scanners/nikto/scanner/Dockerfile
+++ b/scanners/nikto/scanner/Dockerfile
@@ -2,6 +2,25 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-FROM localhost:5000/sullo/nikto
+FROM alpine:3.14 as build
+ARG scannerVersion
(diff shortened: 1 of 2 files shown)


scanners/nikto/scanner/Dockerfile
```diff
@@ -2,6 +2,21 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-FROM localhost:5000/sullo/nikto
+FROM alpine:3.14 as build
+ARG scannerVersion
+RUN apk add git
+RUN git clone https://github.com/sullo/nikto.git /nikto \
+  && cd /nikto \
+  && git switch "nikto-$scannerVersion"
```
SebieF, line 10: One liner alternative: 
```suggestion
RUN git clone --depth 1 --branch "nikto-$scannerVersion" https://github.com/sullo/nikto.git
```

SebieF (approved):

scanners/nikto/scanner/Dockerfile
```diff
@@ -2,6 +2,19 @@
...
+FROM alpine:3.14
+ENV  PATH=${PATH}:/nikto
+COPY --from=build /nikto/program /nikto
 COPY wrapper.sh /wrapper.sh
+RUN apk add --update --no-cache --virtual .build-deps \
+     perl \
+     perl-net-ssleay \
+  && addgroup nikto \
+  && adduser -G nikto -s /bin/sh -D nikto \
+  && chown -R nikto.nikto /nikto
```
J12934, line 19: Is the user not actually used in the image?
Also should the user have a numeric user id?

Is it possible to avoid the chown here by running it during the copy?: https://docs.docker.com/engine/reference/builder/#copy
Otherwise the chowned contents are duplicated in two layers.

SebieF (approved):

PR #650, opened 2021-09-19
Adding integration tests for trivy security scanner

### LINKED ISSUE #623: 🧪 Add Integration Tests for Trivy
The Trivy integration has currently no integration tests. This should be changed so that we can detect issues in the integration early and reliably.

The tests were initially skipped as trivy downloads it's database every time a scan gets executed (unless it's cached), which can lead to flaky test failures.
Ideally when adding the tests a good workaround for this issue can be identified which allows us to load the cache locally so that we don't have to rely on the network connection to the trivy cache service to work during the test.

If not such workaround can be found or is too timeconsuming, it would be ok to add these as is and rely on our existing jest retries to restart false positive test failures.

SebieF:
## Description

Upon merging, this PR will add integration tests for the [trivy](https://github.com/aquasecurity/trivy) security scanner. 
This will close #623.

Contrary to what has been discussed there, downloading the database for trivy during the integration tests does not seem to be an issue. The tests took only about [35s](https://github.com/secureCodeBox/secureCodeBox/runs/3644747912?check_suite_focus=true).
Successful download of the database has also been stated in the trivy logs.

If the tests prove to be unstable, however, it could be considered to add an offline version of the trivy database as described [here](https://aquasecurity.github.io/trivy/v0.18.3/air-gap/). This would pose the need for a scanner dockerfile created by us instead of the official image, where a version of the database would be included. Integration test could be run with the "--skip-update" flag afterwards.  

### Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -976,6 +976,19 @@ jobs:
           cd tests/integration/
           npx jest --ci --color scanner/sslyze.test.js
 
+      # ---- Trivy Integration Tests ----
+
+      - name: "trivy Integration Tests"
+        run: |
+          kubectl -n integration-tests delete scans --all
+          helm -n integration-tests install trivy ./scanners/trivy/  \
+            --set="parser.image.tag=sha-$(git rev-parse --short HEAD)" \
+            --set="parser.image.repository=docker.io/${{ env.DOCKER_NAMESPACE }}/parser-typo3scan" \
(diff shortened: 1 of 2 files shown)


fuhrmeistery (commented):
I'm not quite sure how volatile this test is because of new discovered vulnerabilities. What do you think @J12934 , @rseedorff ?

J12934:
Yeah good point, might be better to have it be a "at least this amount of findings" check, as its bound to go up in the future.

fuhrmeistery (approved):

PR #654, opened 2021-09-20
Automatically Use Correct Batch Apiversion for Nuclei Cronjob

J12934:
Fixes an issue for older Kubernetes versions (1.20 and lower) where the nuclei chart could not be installed correctly.

### DIFF
--- a/scanners/nuclei/templates/nuclei-update-cache-job.yaml
+++ b/scanners/nuclei/templates/nuclei-update-cache-job.yaml
@@ -4,7 +4,7 @@
 
 # We use a persistent volume for central storing of all nuclei-templates to prevent downloading it for each scan again, enabled by default.
 {{ if .Values.nucleiTemplateCache.enabled }}
-{{- if semverCompare ">=1.21-0" .Capabilities.KubeVersion.GitVersion -}}
+{{- if .Capabilities.APIVersions.Has "batch/v1/CronJob" -}}
 apiVersion: batch/v1
 {{- else -}}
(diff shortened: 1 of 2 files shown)


scanners/nuclei/templates/nuclei-update-cache-job.yaml
```diff
@@ -4,7 +4,11 @@
 
 # We use a persistent volume for central storing of all nuclei-templates to prevent downloading it for each scan again, enabled by default.
 {{ if .Values.nucleiTemplateCache.enabled }}
+{{- if semverCompare ">=1.21-0" .Capabilities.KubeVersion.GitVersion -}}
```
rseedorff, line 7:
```suggestion
{{- if semverCompare ">=1.21-0" .Capabilities.KubeVersion.Version -}}
```
Seems to be deprecated ... see https://github.com/helm/charts/issues/20918

rseedorff (changes requested):

scanners/nuclei/templates/nuclei-update-cache-job.yaml
J12934: Ah didn't know about that 👍
`helm create` still outputs a chart which uses the old syntax, I took it from there

J12934:
Also had a issue installing the chart as the activeDeadline setting on `null` on the ScanJob wasn't allowed by a GKE 1.20 apiServer. It's now only acutally set if the value is configured.

rseedorff (approved):

PR #662, opened 2021-09-21
Cache Operator and Lurker Images In CI

fuhrmeistery:
If applied this PR will:
- Add Two Make Targets To Enable Separate Image Builds
- Cache Operator and Lurker Image

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -172,6 +172,7 @@ jobs:
         with:
           name: ${{ matrix.component }}-image
           path: ./operator/${{ matrix.component }}.tar
+          retention-days: 1
 
   # ---- Build Stage | AutoDiscovery ----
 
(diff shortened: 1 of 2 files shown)


J12934:
Maybe we could be "good github citizens" and decrease the retention timespan for the artifacts so github doesn't have to store them for too long: https://github.com/actions/upload-artifact#retention-period

Maybe 3 days? Should be enough for debugging.

fuhrmeistery:
> Maybe we could be "good github citizens" and decrease the retention timespan for the artifacts so github doesn't have to store them for too long: [actions/upload-artifact#retention-period](https://github.com/actions/upload-artifact#retention-period)
> 
> Maybe 3 days? Should be enough for debugging.

Good Point, I think one day could also be sufficient

J12934 (approved):

PR #692, opened 2021-10-01
Add `ftp-anon` and `ftp-banner` nmap scripts and support for portrule scripts

EndPositive:
## Description

This PR, if applied, adds Nmap script support for `ftp-anon` and `ftp-banner`. To achieve this, parser support has been added for portrule scripts. This PR also introduced a generic way for maintainers to add more parsers.

### Context

Nmap has prerule, postrule, hostrule, and portrule [scripts](https://nmap.org/book/nse-script-format.html). Each of these script types output the script result under a different XML field.

For example, the smb-protocols script is a hostrule script and outputs the script results under the `<hostscript>` field within `<hosts>`. However, the ftp-anon scan is a portrule script and will output the script results under the `<port>`  field within <ports>. When parsing and extracting script results, this will need to be taken into account in order to detect whether or not a script was successful and did indeed produce output.

Originally in the SecureCodeBox parser, only the parsing function for one example script was present. This was a hostrule script, and thus got its script results from the <hostscript> subtitle. This caused some confusion when implementing an additional script parsing function, as that script happened to be a portrule script and thus never produced output within <hostscript>, meaning it did not get picked up by the parser as script output.

Pre-rule scripts and post-rule scripts can also have their seperate XML field, namely <prescript> and <postscript> respectively, but these are rarely, if ever, relevant to the creation of a finding for a target host, that special support for these has not been implemented. Note: post-rule scripts often edit output in port-rule or host-rule script fields, which is supported by this merge.

Example host-rule script output
``` xml
<hostscript>
    <script id="smb-protocols" output="&#xa;  dialects: &#xa;    NT LM 0.12 (SMBv1) [dangerous, but default]&#xa;    2.02&#xa;    2.10&#xa;    3.00&#xa;    3.02&#xa;    3.11">
        <table key="dialects">
            <elem>NT LM 0.12 (SMBv1) [dangerous, but default]</elem>
            <elem>2.02</elem>
            <elem>2.10</elem>
            <elem>3.00</elem>
            <elem>3.02</elem>
            <elem>3.11</elem>
        </table>
    </script>
</hostscript>
```

Example port-rule script output
``` xml
<port protocol="tcp" portid="21">
    <state state="open" reason="syn-ack" reason_ttl="0"/>
    <service name="ftp" method="table" conf="3"/>
    <script id="ftp-anon" output="Anonymous FTP login allowed (FTP code 230)"/>
</port>
```

### Checklist

### DIFF
--- a/scanners/nmap/.helm-docs.gotmpl
+++ b/scanners/nmap/.helm-docs.gotmpl
@@ -29,6 +29,22 @@ usecase: "Network discovery and security auditing"
 Nmap ("Network Mapper") is a free and open source (license) utility for network discovery and security auditing. Many systems and network administrators also find it useful for tasks such as network inventory, managing service upgrade schedules, and monitoring host or service uptime.
 
 To learn more about the Nmap scanner itself visit [nmap.org].
+
+## NSE scripts
+
+Currently, the secureCodeBox Nmap parser supports the `smb-protocols`, `ftp-anon`, and `ftp-banner` script with compatibility for hostrules and portrules. If you want custom scripts to be parsed by secureCodeBox, you may contribute your script parser in `parser/parser.js` under `const scriptParser`
+
+```javascript
+const scriptParser = {
+  "ftp-anon": parseFtpAnon,
+  "banner": parseBanner,
+  "smb-protocols": parseSmbProtocols,
+  [...] // Contributed custom parsers
+}
+```
+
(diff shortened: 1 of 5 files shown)


EndPositive:
CodeClimate issues not introduced in this PR.

fuhrmeistery (approved):
Looks great! Thanks for contributing

PR #695, opened 2021-10-01
Add support for hook priorities (ordering)

EndPositive:
## Description
This PR, if applied, schedules hooks based on their priority. Hooks with higher priority are ran before hooks with a lower priority. Hooks with equal priority are scheduled according to the previously defined orderings (i.e. ReadAndWrite hooks first in serial, ReadOnly hooks after in parallel). All hooks have a default priority of `0`.

If current revision is merged, there will be no changes in scheduling behavior yet.

A Documentation PR has also been made https://github.com/secureCodeBox/documentation/pull/131.

### Motivation
The described behavior can be useful when multiple different hooks have been deployed.

Examples:
1. The update-field hook is required to update a field before importing the finding to DefectDojo (R&W).
2. First import finding into DefectDojo (RO) before sending a notification.
3. A data processing hook before any other hook is ran.

### Checklist


operator/apis/execution/v1/scan_types.go
```diff
@@ -101,6 +103,8 @@ type HookStatus struct {
 	HookName string    `json:"hookName"`
 	State    HookState `json:"state"`
 	JobName  string    `json:"jobName,omitempty"`
+	Priority int       `json:"priority"`
```
sanmai-NL, line 106: Why not `uint8`?

operator/config/crd/bases/execution.securecodebox.io_scancompletionhooks.yaml
```diff
@@ -174,6 +178,12 @@ spec:
                       type: string
                   type: object
                 type: array
+              priority:
+                default: 0
+                description: Defines the priority of the hook. Higher priority hooks
```
sanmai-NL, line 183: > Defines the priority of the hook.

I think this sentence is not helpful. Remove it?

Also mention the order of execution within priority classes. (Or reference more detailed docs, but in this case, I would mention this straight away.)

operator/config/crd/bases/execution.securecodebox.io_scans.yaml
```diff
@@ -1775,12 +1775,43 @@ spec:
                       type: string
                     jobName:
                       type: string
+                    priority:
+                      type: integer
+                    state:
+                      description: HookState Describes the State of a Hook on a Scan
```
sanmai-NL, line 1781: End with period, like elsewhere.

operator/crds/execution.securecodebox.io_scancompletionhooks.yaml
```diff
@@ -175,6 +179,12 @@ spec:
                       type: string
                   type: object
                 type: array
+              priority:
+                default: 0
+                description: Defines the priority of the hook. Higher priority hooks
```
sanmai-NL, line 184: I suppose the duplicate text is unavoidable?

operator/controllers/execution/scans/hook_reconciler.go
```diff
@@ -62,250 +60,232 @@ func (r *ScanReconciler) setHookStatus(scan *executionv1.Scan) error {
...
-	switch nonCompletedHook.State {
-	case executionv1.Pending:
-		rawFileURL, err := r.PresignedGetURL(scan.UID, scan.Status.RawResultFile, defaultPresignDuration)
-		if err != nil {
-			return err
-		}
-		findingsFileURL, err := r.PresignedGetURL(scan.UID, "findings.json", defaultPresignDuration)
-		if err != nil {
-			return err
+	if err == nil && scan.Status.State == "ReadAndWriteHookProcessing" {
```
sanmai-NL, line 102: Can you compare to constants (outside this module) rather than literals?

operator/controllers/execution/scans/hook_reconciler.go
```diff
@@ -62,250 +60,232 @@ func (r *ScanReconciler) setHookStatus(scan *executionv1.Scan) error {
...
 
-		rawFileUploadURL, err := r.PresignedPutURL(scan.UID, scan.Status.RawResultFile, defaultPresignDuration)
-		if err != nil {
-			return err
-		}
-		findingsUploadURL, err := r.PresignedPutURL(scan.UID, "findings.json", defaultPresignDuration)
-		if err != nil {
-			return err
-		}
+	if sErr := r.Status().Update(ctx, scan); sErr != nil {
```
sanmai-NL, line 113: I prefer a more descriptive name, like `errorUpdate` or be it `errUpdate`.

operator/controllers/execution/scans/hook_reconciler.go
```diff
@@ -62,250 +60,232 @@ func (r *ScanReconciler) setHookStatus(scan *executionv1.Scan) error {
...
+			priorityQueueItem := util.PriorityQueueItem{
+				Value:    hook,
+				Priority: hook.Priority,
+			}
+			priorityQueue[completedHooks] = &priorityQueueItem
+			completedHooks++
 		}
 	}
+	priorityQueue = priorityQueue[:completedHooks]
+	heap.Init(&priorityQueue)
```
sanmai-NL, line 77: The heap data structure should be an implementation detail of the priority queue (encapsulation).

operator/controllers/execution/scans/hook_reconciler.go
```diff
@@ -62,250 +60,232 @@ func (r *ScanReconciler) setHookStatus(scan *executionv1.Scan) error {
...
+	var err error
+	// If nil then all hooks are done
+	if len(nonCompletedHooks) == 0 {
+		r.Log.Info("Marked scan as done as all hooks have completed", "ScanName", scan.Name)
+		scan.Status.State = "Done"
+		var now = metav1.Now()
+		scan.Status.FinishedAt = &now
+	} else if len(nonCompletedHooks) > 0 {
+		for {
+			priorityQueueItem := heap.Pop(&nonCompletedHooks).(*util.PriorityQueueItem)
```
sanmai-NL, line 135: The source data structure isn't relevant. Can you use `topNonCompletedHook` or some other more relevant name?

(+ Too little encapsulation shows here as well.)

operator/controllers/execution/scans/hook_reconciler.go
```diff
@@ -62,250 +60,232 @@ func (r *ScanReconciler) setHookStatus(scan *executionv1.Scan) error {
...
 		}
 	}
 
-	return nil
-}
-
-func containsJobForHook(jobs *batch.JobList, hook executionv1.ScanCompletionHook) bool {
-	if len(jobs.Items) == 0 {
-		return false
+	if err == nil && scan.Status.State == "ReadOnlyHookProcessing" {
```
sanmai-NL, line 153: Constants.

operator/controllers/execution/scans/hook_reconciler.go
```diff
@@ -62,250 +60,232 @@ func (r *ScanReconciler) setHookStatus(scan *executionv1.Scan) error {
...
+		// Check if there are now still incomplete hooks
+		scan.Status.State = "Completed"
```
sanmai-NL, line 155: Constants.

operator/controllers/execution/scans/hook_reconciler.go
```diff
@@ -62,250 +60,232 @@ func (r *ScanReconciler) setHookStatus(scan *executionv1.Scan) error {
...
+		for _, hook := range scan.Status.ReadOnlyHookStatus {
+			if hook.State != executionv1.Completed {
+				scan.Status.State = "ReadOnlyHookProcessing"
```
sanmai-NL, line 158: Etc.

operator/controllers/execution/scans/hook_reconciler.go
```diff
@@ -62,250 +60,232 @@ func (r *ScanReconciler) setHookStatus(scan *executionv1.Scan) error {
...
+				[]string{
+					rawFileURL,
+					findingsFileURL,
+				},
+			)
+		case executionv1.ReadAndWrite:
+			var rawFileUploadURL string
+			rawFileUploadURL, err = r.PresignedPutURL(scan.UID, scan.Status.RawResultFile, defaultPresignDuration)
+			if err != nil {
+				break
```
sanmai-NL, line 232: The nested breaks are better than the previous duplication, but I propose that you always use switch labels whenever a switch is nested. Just to be very explicit and clear, and avoid defects.

sanmai-NL (changes requested):

operator/controllers/execution/scans/hook_reconciler.go
J12934: This isn't introduced in this PR but something we should definitely do 👍

operator/apis/execution/v1/scan_types.go
J12934: I think having this be a signed int is helpful as this allows you to go both over and under the default prio (0).

J12934:
Thanks for the PR 🚀
Hook Ordering / Prioritization has been a topic we've discussed many times, but never got around to actually implement it.
This already looks conceptually like a good way to tackle the problem.

I'm currently wondering if there is a way to simplify this behavior, as having the Prio applied to both `ReadOnly` and `ReadAndWrite` hooks feels extremely complicated with the differences between serial & parallel execution. 

I'd propose to split the `ReadOnly` hooks in two categories one which runs before `ReadAndWrite` and one which runs after them. (Either configured via a third type, or via a new attribute)

The prio would then only be used for the `ReadAndWrite` hooks and the code would not have to deal (or at least not more then it already does) with orchestrating the hook while respecting both the prio's and the serial / parallel execution order.

EndPositive:
Awesome feedback @J12934! This was the discussion I was looking for :grinning: 

> I'm currently wondering if there is a way to simplify this behavior, as having the Prio applied to both `ReadOnly` and `ReadAndWrite` hooks feels extremely complicated with the differences between serial & parallel execution.

I completely agree that this PR may potentially add unnecessary complexity to the flow of a secureCodeBox scan. I am however, wondering which specific feature from this PR you find adds too much complexity. I'm doubting the flexibility of your proposed solution a bit.

> I'd propose to split the `ReadOnly` hooks in two categories one which runs before `ReadAndWrite` and one which runs after them. The prio would then only be used for the `ReadAndWrite` hooks and the code would not have to deal (or at least not more then it already does) with orchestrating the hook while respecting both the prio's and the serial / parallel execution order.

Adding yet another field (e.g. `order`) for defining the order of `ReadOnly` compared to the `ReadAndWrite` hook adds complexity while reducing the flexibility. As stated in OP, I had a dream for a future implementation without requirements on respective order betwen `ReadAndWrite` and `ReadOnly` hooks. For example, one cannot currently implement a `ReadOnly` -> `ReadAndWrite` -> `ReadOnly` -> `ReadAndWrite` flow. One can also not create a flow of `ReadOnly` -> `ReadOnly` hooks with required orderings (e.g. first upload findings to persistence provider, only then send an email to management with persitence link).

I'm unsure where the complexity you are referring to actually arises. With my initially proposed implementation, the flow is like this:
1. Retrieve all hooks
2. Run all `ReadAndWrite` hooks according to priority. Are **not** allowed to run in parallel.
3. Run all `ReadOnly` hooks according to priority. Are allowed to run in parallel.

For the code, it would be optimal if we disregarded the requirement that `ReadAndWrite` hooks are required to be ran before `ReadOnly` hooks. This logic could simply be implemented with priorities in hook helm charts (i.e. set currently implemented `ReadAndWrite` hook's priority to 1 and `ReadOnly` hooks to 0).

Resulting in the following flow:
1. Retrieve all hooks
2. Run all hooks according to priority. Are allowed to run in parallel if NOT `ReadAndWrite` hook.

The difference between `ReadAndWrite` and `ReadOnly` hooks would simply be their permissions on MinIO while not allowing multiple `ReadAndWrite` hooks simultaneously (preventing possible data loss).

EndPositive:
> Would you then start all ReadOnly Hooks the first ReadAndWrite hook of the same prio at the same time?

Interesting case :thinking: . That would indeed have been undefined behavior in my last proposal.

> Basically the same as it is today just running in multiple "stages" (one "stage" per prio configured, first stage would be all hooks with the highest prio number)

I think that would cover most edges and be as close to what we have today while still really flexible. Sounds good to me!

Shall I update this PR with your latest proposal?

EndPositive:
> @J12934 I'll take a look at failing pipeline

Try again!

sanmai-NL:
My scalability concern was: if you have thousands or more of priority classes, we need an efficient way to process them and the proper, understandable abstraction for such a problem. Whereas you can mimic this with the pre-sorted list and some specific logic to find the top of the list, that would be less standard. I just discussed this with @EndPositive, and it seems our misunderstanding stems from the initial implementation approach with the priority queue. Instead of a priority queue of hooks, a priority queue of classes would've been more suitable. The class struct would contain an ordered list of RW hooks and a list of RO hooks and have a priority (e.g., field). This would be clean, the logic for handling the classes will and should be outside the priority queue implementation. 

However, it seems I overestimated the number of priority classes we foresee. At least, your diagram and the current applications of SCB seem more geared towards a handful of hooks. Please go ahead.

EndPositive:
Just ran into an issue where I wanted to delete a Scan created with pre ordering. I'll take a look.

```
2021-11-01T11:00:17.380Z	ERROR	controller-runtime.manager.controller.scan	Reconciler error	{"reconciler group": "execution.securecodebox.io", "reconciler kind": "Scan", "name": "nmap-example", "namespace": "default", "error": "Scan.execution.securecodebox.io \"nmap-example\" is invalid: [status.readAndWriteHookStatus.priority: Required value, status.readAndWriteHookStatus.type: Required value]"}
github.com/go-logr/zapr.(*zapLogger).Error
	/go/pkg/mod/github.com/go-logr/zapr@v0.2.0/zapr.go:132
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.8.3/pkg/internal/controller/controller.go:302
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.8.3/pkg/internal/controller/controller.go:253
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func1.2
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.8.3/pkg/internal/controller/controller.go:216
k8s.io/apimachinery/pkg/util/wait.JitterUntilWithContext.func1
	/go/pkg/mod/k8s.io/apimachinery@v0.20.2/pkg/util/wait/wait.go:185
k8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1
	/go/pkg/mod/k8s.io/apimachinery@v0.20.2/pkg/util/wait/wait.go:155
k8s.io/apimachinery/pkg/util/wait.BackoffUntil
	/go/pkg/mod/k8s.io/apimachinery@v0.20.2/pkg/util/wait/wait.go:156
k8s.io/apimachinery/pkg/util/wait.JitterUntil
	/go/pkg/mod/k8s.io/apimachinery@v0.20.2/pkg/util/wait/wait.go:133
k8s.io/apimachinery/pkg/util/wait.JitterUntilWithContext
	/go/pkg/mod/k8s.io/apimachinery@v0.20.2/pkg/util/wait/wait.go:185
k8s.io/apimachinery/pkg/util/wait.UntilWithContext
	/go/pkg/mod/k8s.io/apimachinery@v0.20.2/pkg/util/wait/wait.go:99

```

EndPositive:
@J12934, just pushed a fix for Scan objects which were marked `Done` and were created before this change. Once the operator detected deletion of a Scan resource, it tried to remove the finalizer, which require that the `ReadAndWriteHookStatus` was valid. I.e. we have to check whether the field had already been migrated on deletion.

EndPositive:
ab00c6c  added hook prio field in values.yaml

J12934 (approved):

malexmave:
Hi @EndPositive. I am ready to merge this, but after merging your other hook PR, this is giving some conflicts. All the values.yaml changes are easy to resolve, but the conflicts in the hook_reconciler are complex enough that I would prefer if you could quickly resolve them, so that I don't introduce any new bugs in the process. Can you quickly resolve them?

EndPositive:
Done. I think you should run integration test & helm-docs again.

J12934 (approved):

PR #697, opened 2021-10-04
fix: update npm deps ./bin/npm-audit-fix-all.sh

fuhrmeistery:
Signed-off-by: Yannik Fuhrmeister <yannik.fuhrmeister@iteratec.com>

### DIFF
--- a/hooks/cascading-scans/hook/package-lock.json
+++ b/hooks/cascading-scans/hook/package-lock.json
@@ -1315,9 +1315,9 @@
       }
     },
     "node_modules/ansi-regex": {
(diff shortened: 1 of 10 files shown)


SebieF (commented):
The package.json files have to be updated as well, don't they?

fuhrmeistery:
> The package.json files have to be updated as well, don't they?

Good question, but the package.json itself did not change because these changes only affect sub packages or are already included in the current version spec.

J12934 (approved):

J12934 (approved):

PR #714, opened 2021-10-08
:construction: [Consistency] Making all values.yaml files consistent

### LINKED ISSUE #693: :construction: [Consistency] Make all values.yaml files follow the same template
Part of #620.
As a secureCodeBox developer I would like to have all Chart.yaml files consistent, so that maintenance and development gets simplified.

SebieF:
Closes #693.

Once applied, all values.yaml files for all scanners will follow the same template. The following changes have been made:

1. Replaced some unnecessary comments
2. Added xy.xyz.zxy descriptions to all values
3. Added **security context** to all scanners
4. Added cascading rules to all scanners (when no cascading rules exist, they are set to false)
5. Changed nuclei image tag from "latest" to "null"
6. Set RunAsNonRoot for whatweb from "false" to "true"
7. ncrack, zap/zap-advanced have some special settings that have not been touched

Note that in the amass values.yaml securityContext, runAsNonRoot had to be set to false, because their [Dockerfile](https://github.com/OWASP/Amass/blob/master/Dockerfile) does not use a numeric value for their user. 
Thus their user cannot be verified to be root or not:
Error: container has runAsNonRoot and image has non-numeric user (user), cannot verify user is non-root (pod: "scan-amass-scanner-dummy-scan, container: amass). (See #715)

Same unfortunately applies for even more scanners, see #723 

### Checklist


SebieF:
We discussed the problem and it was concluded that the correct version should be used at the moment, but an issue for the inconsistent versioning has been created (#716).

Issue #723, opened 2021-10-13
Scanners use runAsNonRoot: false in their values.yaml securityContext


SebieF:
Once PR #714 is applied, all values.yaml files have a securityContext set. During testing this, it was noted that some scanners have to use *runAsNonRoot: false* and/or *readOnlyRootFilesystem: false* in order to function. This can be due to the behaviour of the underlying scanner or because the Dockerfile uses a non-numeric user. Fixing this may require our own Scanner Dockerfile where an already build original image is used or pull requests in other repositories in order to have them use non-numeric, kubernetes-friendly users.

See also: #285 

Affected scanners:
- [ ] amass (#715)
- [ ] cmseek (readOnlyRootFilesystem)
- [ ] gitleaks
- [ ] kube-hunter
- [ ] ssh-scan
- [ ] sslyze
- [ ] trivy
- [ ] typo3scan (readOnlyRootFilesystem)
- [ ] wpscan
- [ ] zap
- [ ] zap-advanced

Weltraumschaf:
Relates to https://github.com/secureCodeBox/secureCodeBox/issues/1330

PR #727, opened 2021-10-13
:construction: [Consistency] Make all files of cmseek follow consistent templates

SebieF:
## Description

Because cmseek was added as a new security scanner while recent consistentcy refactoring was done, some files from cmseek do not follow the new template structure. This PR fixes all the inconsistencies with other scanners. 
It also seems like the securityContext of values.yaml needs readOnlyRootFilesystem to be false (see #723).

### Checklist


J12934:
@SebieF this seems to include a lot of changes from the other consistency pr's not just the cms seek

PR #804, opened 2021-11-10
Add ability to configure affinity and tolerations for scans and hooks

### LINKED ISSUE #803: Use preemptible node for scans
## Your Question

Currently, I am using preemptible nodes on my cluster to save costs, so I would like scans to be deployed on preemptible nodes. For secureCodeBox operator, I will deploy on a normal node (non-preemptible).

I tried apply 1 configuration as follows, but it doesn't work.
```
kind: Scan
metadata:
  name: "trivy-mediawiki"
spec:
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: cloud.google.com/gke-nodepool
            operator: In
            values:
            - preemptible
  tolerations:
  - key: "dedicated"
    operator: "Equal"
    value: "preemptible"
    effect: "NoSchedule" 
    
  scanType: "trivy"
  parameters:
    - "mediawiki:stable"
```
```
❯ k apply -f preemptible-trivy-test.yaml
error: error validating "preemptible-trivy-test.yaml": error validating data: [ValidationError(Scan.spec): unknown field "affinity" in io.securecodebox.execution.v1.Scan.spec, ValidationError(Scan.spec): unknown field "tolerations" in io.securecodebox.execution.v1.Scan.spec]; if you choose to ignore these errors, turn validation off with --validate=false
```

**So is there any way I can do this?**

malexmave:
## Description
This PR resolves #803 by adding affinity and tolerations to the ScanSpec and ScanCompletionHookSpec.

Currently in draft state. Tasks:

- [x] Add to `ScanSpec` and `ScanCompletionHookSpec` CRDs and re-generate them
- [x] Add to controller (pull information from CRD and put it into the pod)
- [ ] Check if the JS definitions need to be updated (hooks, SDKs)
- [ ] Add new inheritX labels
- [ ] Add test cases for cascading hooks
- [ ] Test everything
- [ ] Double-check if the docs need to be updated anywhere or if everything is done automatically anyway
- [ ] @J12934: Is this something that would be useful to also add to the values.yaml for the different scanners? That way, you could have specific scan types always run with a specific toleration / affinity and would not need to specify it every single time.

I will try to wrap this up later today so we can release it as part of 3.4 - that way the inconvenience of having to manually update CRDs can be lessened, as a CRD update is necessary anyway because of the new hook features. (Not looking forward to the merge conflicts - perhaps we should merge the other two PRs so I can quickly rebase on that branch to avoid that pain down the road?)

### DIFF
--- a/hooks/cascading-scans/README.md
+++ b/hooks/cascading-scans/README.md
@@ -162,6 +162,7 @@ zap-http         zap-baseline-scan   non-invasive   medium
 |-----|------|---------|-------------|
 | hook.image.repository | string | `"docker.io/securecodebox/hook-cascading-scans"` | Hook image repository |
 | hook.image.tag | string | defaults to the charts version | The image Tag defaults to the charts version if not defined. |
+| hook.labels | object | `{}` | Add Kubernetes Labels to the hook definition |
 | hook.priority | int | `0` | Hook priority. Higher priority Hooks are guaranteed to execute before low priority Hooks. |
 | hook.ttlSecondsAfterFinished | string | `nil` | Seconds after which the kubernetes job for the hook will be deleted. Requires the Kubernetes TTLAfterFinished controller: https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/ |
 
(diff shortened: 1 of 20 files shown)


malexmave:
Okay, I broke something with my attempts to get this back on top of main, and even force-pushing this branch did not fix it. Since there isn't a lot of interesting history yet, I'll close this PR and open a new one that is less borked.

PR #808, opened 2021-11-11
Adds Notice File Which is Required by ASF 2.0

Weltraumschaf:
As described in https://infra.apache.org/licensing-howto.html this
adds a minimal notice file.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,0 +1,5 @@
+secureCodeBox
+Copyright 2020 iteratec GmbH
+
+This product includes software developed at
+iteratec GmbH (http://www.iteratec.com/).


Weltraumschaf:
I'm not sure if we should add Secura in this file like:

> Portions of this software were developed at
> Secura (https://www.secura.com).

J12934:
I'm confused why this would be required for us.

The link you posted (https://infra.apache.org/licensing-howto.html) states that is required for Apache Software Foundation projects but not that it is required by projects using the Apache 2.0 license. 
The NOTICE part of the license starts with `If the Work includes a "NOTICE" text file as part of its distribution...` which really doesn't sound like it is required.

Weltraumschaf:
As I read the docs: For ASF projects it is mandatory to have this file. For non-ASF projects it should be present. AFAIU this file is especially if a contributor wants to be notices explicitly. That's what Thomas (our license expert) told me.

The [section 4](https://www.apache.org/licenses/LICENSE-2.0.html#redistribution) of the license is only relevant for others doing derivative work of this distribution.

At the end i'm not 100 % sure if we need this.I try to find more information.

Weltraumschaf:
In [this Stack Exchange answer](https://opensource.stackexchange.com/a/8567) it is explained for what this file is good for:

> the NOTICE should be used to identify the following information:
>
> * The copyright holders (and potentially patent holders)
> * The portions of the source code for which they hold the copyright (and that potentially can be covered by their patents)
> * The License adopted for that portion of the code

AFAIR we have some files (some examples I guess) in our repo with different licenses. So we should mention this in the NOTICE file.

thelbich:
> I'm confused why this would be required for us.
> 
> The link you posted (https://infra.apache.org/licensing-howto.html) states that is required for Apache Software Foundation projects but not that it is required by projects using the Apache 2.0 license. The NOTICE part of the license starts with `If the Work includes a "NOTICE" text file as part of its distribution...` which really doesn't sound like it is required.

We have to distinguish between the Apache 2.0 license (included in the license file) and copyright issues. The copyright information (e.g. (c) 2015 by ..., if contributors wanted to be listed explicitly) has to be collected in the NOTICE file. If nobody claims copyrights of parts of the software, a notice file is not necessary (that's why is is stated as optional in the Apache license), but in practice, most bigger software projects (like the SCB) normally includes copyright statements (collected in the notice file).

Weltraumschaf:
@thelbich Thanks for your answer.

Three questions:

1. What about files we have in the repo with different licenses? The Stack Exchange answer I linked states that we should add these different licenses in the notice file too.
2. What about the contribution from Secura? Should we mention them in the notice file?
3. Does someone really insist to be noticed, if we enforce them to sign-off their contribution? AFAIU ist the whole DCO and sign-off thing we do that every contributor give up her copyright.

sanmai-NL:
The contributions by Secura should have the proper SPDX headers. :slightly_smiling_face: 
What is the relationship with DCO?

Weltraumschaf:
Problem is: we are all no lawyers :-| I asked the FSFE months ago for help, but the only answer I got was "it's quite complicated, forwarding to our lawyers..." No answer yet. I try to contact someone at ASF. Maybe they can clarify this question.

Weltraumschaf:
I [filed an issue|https://issues.apache.org/jira/browse/LEGAL-587] at ASF legal affairs.

Weltraumschaf:
The ASF answered my question and closed the issue. The answer is: You may or may not add a NOTICE file and read section 4d of the licens 🤷‍♂️ Neither helpful nor kind 😕

sanmai-NL:
Then let's not do it, unless it supports automation (like SPDX does).

malexmave:
Agreed. If we don't know what we would have to include in the file, but are pretty sure that we don't actually need the file at all, we save ourselves some headache by just leaving it out.

PR #839, opened 2021-11-25
Upgrading Nmap from 7.91-r0 to 7.92-r2

EndPositive:
Release changes can be found here: https://nmap.org/changelog.html

There are small differences in XML output, though we don't explicitly parse them. I've updated the XML's which I could reproduce and the unit tests still pass.

Note that the base image has also been updated from alpine 3.13 to 3.15 so that this Nmap version can be installed from the default repo.


Ilyesbdlala (approved):
Looks good to me.

Issue #834, opened 2021-11-24
Update list of scanners for telemetry


malexmave:
## 🐞 Bug report
As a SCB developer, I would like to know which scanners are used how often, so that we can prioritize which to support and consider removing unused scanners that cause a lot of maintainance work.

Telemetry is only sent for scanners on an allowlist (as we do not want to leak internal information in case people are developing their own scanners). However, this allowlist has not been updated in a while, and is thus missing some scanners. It will have to be updated [both in this repo](https://github.com/secureCodeBox/secureCodeBox/blob/c4275e9495f783c160157475157d986a9e5f31e9/operator/internal/telemetry/telemetry.go#L26) and in the [telemetry repo](https://github.com/secureCodeBox/telemetry).

PR #846, opened 2021-11-29
Update list of scantypes for telemetry

malexmave:
Part of #834 - updating the list of official scantypes in the telemetry so we can get a better idea of which scanners are actually being used, so that we can prioritize supporting these scanners as opposed to other scanners that are never actually used.


malexmave:
https://github.com/secureCodeBox/secureCodeBox/blob/4586eca3b6bc8676c76186396397a42957dae390/operator/internal/telemetry/telemetry.go#L46-L49

Are these still correct? In the scantypes we only have `zap` and `zap-advanced` as scantype names, but I am not sure if ZAP gets special treatment for a reason here.

EndPositive:
@malexmave zap has 3 scantypes declared in its `zap-scan-type.yaml`:

https://github.com/secureCodeBox/secureCodeBox/blob/ae7a050470a586344edc832001a38c2c0bbc962d/scanners/zap/templates/zap-scan-type.yaml#L8

https://github.com/secureCodeBox/secureCodeBox/blob/ae7a050470a586344edc832001a38c2c0bbc962d/scanners/zap/templates/zap-scan-type.yaml#L56

https://github.com/secureCodeBox/secureCodeBox/blob/ae7a050470a586344edc832001a38c2c0bbc962d/scanners/zap/templates/zap-scan-type.yaml#L99

the `zap-advanced` scantype is created in the `zap-advanced` scanner

https://github.com/secureCodeBox/secureCodeBox/blob/ae7a050470a586344edc832001a38c2c0bbc962d/scanners/zap-advanced/templates/zap-advanced-scan-type.yaml#L21

malexmave:
Ah, I see, perfect. In that case this should be ready to go, but it's probably better to wait for an expert for the telemetry code to review this - it's not that urgent anyway :).

Weltraumschaf (approved):
LGTM

J12934 (approved):

Issue #286, opened 2021-02-01
Dedicated User For Scanner Images


fuhrmeistery:
Affected Scanner Images:
- [ ] angular-csti-scanner #852 
- [x] gitleaks #526 
- [x] git-repo-scanner #527 
- [x] kube-hunter #836 
- [x] ncrack #560 
- [x] nikto #642 
- [x] screenshooter #559

See #285

PR #852, opened 2021-12-01
Run angularjs-csti As User In Container

Ilyesbdlala:
angularjs-csti scanner now runs as NonRoot
* A new user is added to the docker image
* Path to wrapper.sh is changed to new user's
home directory
It should be noted that angularjs-csti currently does not have any integration tests. So testing was done manually.
See: #286

### DIFF
--- a/scanners/angularjs-csti-scanner/scanner/Dockerfile
+++ b/scanners/angularjs-csti-scanner/scanner/Dockerfile
@@ -4,8 +4,18 @@
 
 FROM python:3.6-alpine
 ARG scannerVersion
-COPY acstis-script.py /acstis/acstis-script.py
-COPY wrapper.sh /wrapper.sh
-RUN apk add --update --no-cache g++ gcc libxslt-dev
-RUN pip install https://github.com/tijme/angularjs-csti-scanner/archive/$scannerVersion.zip
-ENTRYPOINT [ "sh", "/wrapper.sh" ]
+
+RUN apk add --update --no-cache g++ gcc libxslt-dev \
+    && pip install https://github.com/tijme/angularjs-csti-scanner/archive/$scannerVersion.zip
+
+RUN adduser -S -H -u 1001 angularjscsti
+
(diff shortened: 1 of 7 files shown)


scanners/angularjs-csti-scanner/scanner/Dockerfile
```diff
@@ -4,8 +4,18 @@
...
+
+RUN apk add --update --no-cache g++ gcc libxslt-dev \
+    && pip install https://github.com/tijme/angularjs-csti-scanner/archive/$scannerVersion.zip
+
+RUN adduser -S -H -u 1001 angularjscsti
+
+COPY acstis-script.py /home/angularjscsti/acstis/acstis-script.py
+COPY wrapper.sh /home/angularjscsti/wrapper.sh
+
+RUN pip install https://github.com/tijme/angularjs-csti-scanner/archive/$scannerVersion.zip \
```
malexmave, line 16: Isn't this redundant (installing the same pip package twice)?

Ilyesbdlala: yes :sweat_smile: . It's a copy paste mistake. I'll resolve it.

malexmave:
@Ilyesbdlala can you double-check if the documentation requires updates as well because the paths changed? From a quick look, it seems like we have some references to absolute paths [in the docs](https://docs.securecodebox.io/docs/scanners/angularjs-csti-scanner#request-configuration) that may require changing.

Ilyesbdlala:
This PR also sets readOnlyRootFileSystem in values.yaml is to false to allow the wrapper.sh to write to the config file.
This is for now required to enable using external configs.
It should be noted that the angularjs-csti is currently borked in the main repo. This is due to `runAsNonRoot` and `readOnlyRootFilesystem` both being set to true. This PR allows the scanner to run as non root.
See also : https://github.com/secureCodeBox/secureCodeBox/issues/723

PR #855, opened 2021-12-01
Update Version Tags

susuproject:
Sometimes I get an image pull error. Using the latest tag may be more effective.

### Checklist

### DIFF
--- a/operator/Chart.yaml
+++ b/operator/Chart.yaml
@@ -9,7 +9,7 @@ description: secureCodeBox Operator to automate the execution of security scans 
 type: application
 
 # version - gets automatically set to the secureCodeBox release version when the helm charts gets published
-version: v3.1.0-alpha1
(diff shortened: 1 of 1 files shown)


malexmave:
Hi @susuproject. The tags are set as part of the release process, so whatever is written in these files here only has an influence if you do not install the release versions of the charts from our helm repository, and instead install them based on what is inside this repo directly (which we do not recommend).

If you are installing from our helm repository, you should not be getting any image pull errors - can you reproduce the error, and if so, are you trying with the latest version? Is there a more specific error message in the k8s logs that tells you what exactly about the image pull is failing?

malexmave:
Dear @susuproject, since we didn't hear back from you, I'll go ahead and close this PR. But please feel free to open a new issue if you can reproduce the problem!

PR #858, opened 2021-12-06
Fix/nuclei without template cache pvc

J12934:
## Description

Related to secureCodeBox/documentation#166

Noticed that there was still a bug in the chart when disabling the template cache.
Also added docs on how to actually disable it.

### Checklist

### DIFF
--- a/scanners/nuclei/.helm-docs.gotmpl
+++ b/scanners/nuclei/.helm-docs.gotmpl
@@ -137,6 +137,19 @@ STATISTICS:
 {{- end }}
 
 {{- define "extra.chartConfigurationSection" -}}
+## Install Nuclei without Template Cache CronJob / PersistentVolume
+
+Nuclei uses dynamic templates as its scan rules, these determine which requests are performed and which responses are considered to be a finding.
+These templates are usually dynamically downloaded by nuclei from GitHub before each scan. When you are running dozens of parallel nuclei scans you quickly run into situations where GitHub will rate limit you causing the scans to fail.
+To avoid these errors we included a CronJob which periodically fetches the current templates and writes them into a kubernetes PersistentVolume (PV). This volume is then mounted (as a `ReadOnlyMany` mount) into every scan so that nuclei scans have the up-to-date templates without having to download them on every scan.
+
+Unfortunately not every cluster supports the required `ReadOnlyMany` volume type.
+In these cases you can disable the template cache mechanism by setting `nucleiTemplateCache.enabled=false`.
+Note thought, that this will limit the number of scans you can run in parallel as the rate limit will likely cause some of the scans to fail.
+
+```bash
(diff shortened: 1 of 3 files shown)


malexmave:
Did you also check the rest of the Nuclei documentation (since you are closing the general issue for sanity-checking the docs with this PR), or is this only for the one specific problem related to template caches (in which case we should leave the issue open)?

J12934:
> Did you also check the rest of the Nuclei documentation (since you are closing the general issue for sanity-checking the docs with this PR), or is this only for the one specific problem related to template caches (in which case we should leave the issue open)?

Ah nope, good point 👍

Ilyesbdlala:
installing nuclei with the following works. templates are downloaded.
```
helm install nuclei secureCodeBox/nuclei --set="nucleiTemplateCache.enabled=false"
```
However running it normally (with the template cache ) results in forever `pending` pods. Does that mean that my cluster created with "kind"  simply doesn't support the `ReadOnlyMany` volume type you mention. Or is this not the expected behaviour? 
```
NAME                                          READY   STATUS    RESTARTS   AGE  
nuclei-update-initial-template-cache-2lrnt    0/1     Pending   0          74m  
nuclei-update-template-cache-27316200-h6dbz   0/1     Pending   0          8m46s  
```

J12934:
yup probably, you can check by running 

```bash
kubectl describe nuclei-content-claim
```

This should include a error description in the pvc.

Note: the 3.4.0 helm chart currently also had some issue with the security context, might also be related and blocking your containers from being executed

Ilyesbdlala:
Just for reference, the command the view the PersistentVolumeClaim was 
`
kubectl describe pvc nuclei-content-claim
`
And yes, my cluster unfortunately doesn't support `ReadOnlyMany` volumes. 
```
Type     Reason                Age                    From                                                                                                Message
  ----     ------                ----                   ----                                                                                                -------
  Normal   WaitForFirstConsumer  28m                    persistentvolume-controller                                                                         waiting for first consumer to be created before binding
  Normal   Provisioning          12m (x7 over 28m)      rancher.io/local-path_local-path-provisioner-547f784dff-cw2sb_befcd024-db22-44bd-a681-d8162826802c  External provisioner is provisioning volume for claim "default/nuclei-content-claim"
  Warning  ProvisioningFailed    12m (x7 over 28m)      rancher.io/local-path_local-path-provisioner-547f784dff-cw2sb_befcd024-db22-44bd-a681-d8162826802c  failed to provision volume with StorageClass "standard": Only support ReadWriteOnce access mode
  Normal   ExternalProvisioning  3m10s (x103 over 28m)  persistentvolume-controller                                                                         waiting for a volume to be created, either by external provisioner "rancher.io/local-path" or manually created by system administrator
```
works well otherwise.

Ilyesbdlala (approved):

PR #866, opened 2021-12-08
Add k8s 1.23 to CI Matrix and Update Existing Versions

J12934:
Updates the list of kubernetes versions used in ci to the include the new 1.23 release and removes the old 1.19 version which we don't officially support anymore.

After this is merged we have to update the k8s versions referenced on our docs site.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -703,10 +703,10 @@ jobs:
     strategy:
       matrix:
         k8sVersion:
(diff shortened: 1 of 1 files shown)


J12934:
Note the required integration tests also need to be updated the new version numbers thats why there are expected but not executed integration test checks displayed for this pr

Weltraumschaf (approved):
LGTM

PR #903, opened 2021-12-21
Git Repo Scanner: Add repository topics to output

### LINKED ISSUE #893: git-repo-scanner: Add repo topics to output
~As a security team member, I want to select only specific repos for scanning, e.g. exclude those with a certain tag, or only scan those with a certain tag.~
As a security scanner, I want to be able to trigger cascading scans based on repository topics, so I need information about the topics in the output of git-repo-scanner.

~Add switches to include/exclude certain tags.~
Include information about repository topics in the output of git-repo-scanner.

Keep everything as-is. 😁

malexmave:
Closes #893 by adding repository topics to the output of the repo scanner. The repositories are shown as a list, so natively, it will be a bit difficult to filter based on them in the cascading scans. However, it should be possible to include/exclude repositories with specific tags from cascading scans using the scope limiter functionality that was recently introduced by @EndPositive. I will do some testing with that and add a bit of documentation describing how to do that once I have a good solution.

This change is "API-budget neutral" for GitLab, but currently causes GitHub scans to consume twice the number of API calls because the PyGithub library currently doesn't expose the list of topics without making another call, even though the API returns it on the first call as well. I have [opened an issue in that repo](https://github.com/PyGithub/PyGithub/issues/2131) - if it gets resolved, we can update the library and go back to using only a single API call per repo. As I hope that this is temporary, and to avoid inconsistencies with the GitLab scanner, I did not add a flag to enable / disable the topic lookup - but if you think that it would be a good idea to do so, I can certainly add it.


Weltraumschaf (approved):
LGTM, but some words in the commit message why it is done like it is would be helpful to understand the changes.

Issue #454, opened 2021-05-27
Add a persistence Hook to send Findings to Azure Log Analytics / Azure Sentinel


rseedorff:
### Is your feature request related to a problem?

As a secureCodeBox user I would like to analyse the secureCodeBox findings with Azure Log Analytics / Azure Sentinel. This can be seen as an alternative to the existing elasticsearch or defecdojo integration hooks (persistence provider).

### Describe the solution you'd like

With the Azure HTTP Data Collector API it is possible to retrieve JSON based log data which can be used by Azure Log Analytics / Azure Sentinel. There are several existing documentations how to implement a client in python or even nodejs. It should be easy to integrate within the secureCodeBox as persistence hook.

### Additional context

- https://docs.microsoft.com/en-us/azure/azure-monitor/logs/data-collector-api
- https://docs.microsoft.com/en-us/azure/azure-monitor/logs/data-collector-api#python-3-sample
- https://securecloud.blog/2019/12/20/nodejs-logging-integration-with-azure-log-analytics-sentinel/
- https://github.com/sportsmgmt-labs/Azure-Log-Analytics-Node-Function/blob/master/index.js
- https://github.com/Azure/Azure-Sentinel

PR #913, opened 2022-01-06
Add Azure Monitor persistence hook

malexmave:
This PR adds the Azure Monitor persistence provider, as requested in #454.

## Description
Azure Monitor is the logging service by Microsoft. It allows configuring alerting based on events in logs, and also serves as the basis for Microsoft Sentinel, the cloud SIEM by Microsoft. As such, it can be desireable to push results from SCB into the system so that they can be acted upon.

This hook uses the [data collector API](https://docs.microsoft.com/en-us/azure/azure-monitor/logs/data-collector-api) to push the data into custom log types (one per scan type). For more details, read the included documentation.

At the moment, the code is fairly bare-bones and does not check if the data actually conforms to the [expectations that Azure Monitor has](https://docs.microsoft.com/en-us/azure/azure-monitor/logs/data-collector-api#data-limits), namely, a maximum of:
- 30 MB per POST request
- 32 KB per individual field value
- 50 characters per column name

The consequences for not respecting the field value and column name size limitations are truncation of the data. Too large POST requests will likely be rejected by the system with a HTTP error (not explicitly specified in docs). If desired, I can add more checks to the code to validate these requirements explicitly - I expect that 99.999% of secureCodeBox findings will fall within them without any extra modification. Truncation seems like an okay outcome to me (and cannot be prevented using checks on our end, since the only thing we could do is to truncate client-side). So the only place where we can actually do anything useful based on the client-side checks would be to split up POSTs that are larger than 30 MB into multiple individual POSTs. Do we expect findings to be larger than 30 MB?

### Checklist


hooks/persistence-azure-monitor/.helm-docs.gotmpl
```diff
@@ -0,0 +1,59 @@
...
+```bash
+# Write the secrets to a file without trailing linebreak
+# (use leading space to exclude it from the history of your shell)
+ echo -n your-workspace-id > workspace
+ echo -n your-shared-key > sharedkey
+# Create the secret
+kubectl create secret generic azure-monitor --from-file=workspace=./workspace --from-file=sharedkey=./sharedkey
+# Delete the files again
+rm workspace sharedkey
+```
```
J12934, line 48: Why does this have to be file?

Could they not be created using `--from-literal`?

hooks/persistence-azure-monitor/Chart.yaml
```diff
@@ -0,0 +1,28 @@
...
+apiVersion: v2
+name: persistence-azure-monitor
+description: The Azure Monitor persistence provider persists secureCodeBox findings into Azure Monitor.
+
+type: application
+
+# version - gets automatically set to the secureCodeBox release version when the helm charts gets published
+version: v3.1.0-alpha1
+
+appVersion: 7.9.2
```
J12934, line 14: Where does 7.9.2 come from?

hooks/persistence-azure-monitor/hook/hook.js
```diff
@@ -0,0 +1,89 @@
...
+  scan,
+  workspaceId = process.env["MONITOR_WORKSPACE_ID"],
+  sharedKey = process.env["MONITOR_SHARED_KEY"],
+  logTypePrefix = process.env["MONITOR_LOGTYPE_PREFIX"],
+  crypto = require("crypto"),
+  fetch = require('node-fetch')
+}) {
+  if (!(workspaceId && sharedKey)) {
+    console.error("Missing Workspace ID or shared key. Please provide them in the MONITOR_WORKSPACE_ID and MONITOR_SHARED_KEY environment variables");
+    return 1;
```
J12934, line 20: The return value does not do anything here as it isn't intepreted by the hook-sdk, see: https://github.com/secureCodeBox/secureCodeBox/blob/9d4e49a3f71fe96709bbf43b75e43aafdd8c556d/hook-sdk/nodejs/hook-wrapper.js#L154

You can exit directly using process.exit(1) or throwing an error

hooks/persistence-azure-monitor/hook/package.json
```diff
@@ -0,0 +1,46 @@
...
+            "name": "Max Maass",
+            "url": "https://github.com/malexmave"
+        }
+    ],
+    "bugs": {
+        "url": "https://github.com/secureCodeBox/secureCodeBox/issues"
+    },
+    "license": "Apache-2.0",
+    "dependencies": {
+        "crypto": "^1.0.1",
```
J12934, line 38: Crypto is now build in nodejs :) The dependency can & should be removed, see: https://www.npmjs.com/package/crypto

hooks/persistence-azure-monitor/hook/package.json
```diff
@@ -0,0 +1,46 @@
...
+        "fetch": "^1.1.0",
```
J12934, line 39: This import doesn't seem to be used anymore?

hooks/persistence-azure-monitor/hook/package.json
```diff
@@ -0,0 +1,46 @@
...
+        "node-fetch": "2.0"
+    },
+    "devDependencies": {
+        "jest": "^27.0.6",
+        "jest-fetch-mock": "^3.0.3"
```
J12934, line 44: Also unused?

hooks/persistence-azure-monitor/hook/hook.js
```diff
@@ -0,0 +1,89 @@
...
+  const signature = crypto.createHmac('sha256', Buffer.from(sharedKey, 'base64')).update(headersToAuthenticate, 'utf-8').digest('base64');
+
+  // Create the header in the correct form
+  const auth = 'SharedKey ' + workspaceId + ':' + signature;
+  // Create the header dictionary
+  const headers = {
+    "content-type": "application/json", 
+    "Authorization": auth,
+    "Log-Type": logType,
+    "x-ms-date": processingDate
```
J12934, line 73: Is this mix between different capitalization styles required here?

hooks/persistence-azure-monitor/hook/hook.test.js
```diff
@@ -0,0 +1,83 @@
+// SPDX-FileCopyrightText: 2022 iteratec GmbH
+//
+// SPDX-License-Identifier: Apache-2.0
+
+process.env = Object.assign(process.env, {"MONITOR_WORKSPACE_ID": '123123', 
+                                          "MONITOR_SHARED_KEY": "ffffffffffffffffffffffffffffff",
+                                          "MONITOR_LOGTYPE_PREFIX": "SCB"});
```
J12934, line 7: Seems simpler to me :)

```suggestion
process.env["MONITOR_WORKSPACE_ID"] = "123123";
process.env["MONITOR_SHARED_KEY"] = "ffffffffffffffffffffffffffffff";
process.env["MONITOR_LOGTYPE_PREFIX"] = "SCB";
```

hooks/persistence-azure-monitor/hook/hook.test.js
```diff
@@ -0,0 +1,83 @@
...
+
+  await handle({getFindings,
+                scan,
+                workspaceId: "123123",
+                sharedKey: "ffffffffffffffffffffffffffffff",
+                logTypePrefix: "SCB",
+                fetch: fetch,
+                });
+
+  expect(fetch).toBeCalledTimes(1);
```
J12934, line 54: Can we also add asserts to what fetch was called with?
To ensure the header auth mechanisms work correctly and if the findings are properly enriched / serialized

J12934 (dismissed):
Looks great 😊
Noticed a few things while going over the code

hooks/persistence-azure-monitor/hook/hook.js
malexmave: I'm not sure, but it is how it is done in the [official examples](https://docs.microsoft.com/en-us/azure/azure-monitor/logs/data-collector-api#sample-requests), so I would maintain it, just in case.

hooks/persistence-azure-monitor/hook/hook.js
```diff
@@ -0,0 +1,110 @@
+// SPDX-FileCopyrightText: 2022 iteratec GmbH
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// Fixed settings for the script
+const monitorApiVersion = "2016-04-01";
+const processingDate = new Date().toUTCString();
```
J12934, line 7: This is now duplicated
```suggestion

```

rseedorff (dismissed):

rseedorff (approved):

PR #916, opened 2022-01-06
Update to JuiceShop 13.0.3

J12934:


malexmave:
Is this a regular small version bump, or do we expect that this may require changes to the documentation (e.g. when talking about the expected findings in documentation examples that use Juice Shop)?

J12934:
@malexmave 
No major changes to the challenges / vulns which would change the scan results.
https://github.com/juice-shop/juice-shop/releases/tag/v13.0.0

malexmave (approved):

PR #923, opened 2022-01-11
Fix logging of cascading scan name

### LINKED ISSUE #921: Cascading Scan hook logs "undefined" scan name
With SCB Operator and cascading scan hook v3.6.0, the cascading scan logs show a lot of "undefined" scan names when starting new scans:

```
Fetched 15 findings from the file storage
Fetching CascadingScans using LabelSelector: "securecodebox.io/intensive=light,securecodebox.io/invasive=non-invasive"
Fetched 1 CascadingRules
Starting Scan undefined
Starting Scan undefined
Hook completed
```

It seems like the hook only [sets the `scan.metadata.generateName`](https://github.com/secureCodeBox/secureCodeBox/blob/e46688ecd2c71697a83e93c972e549847977679a/hooks/cascading-scans/hook/hook.ts#L164), but later tries to [read the `scan.metadata.name`](https://github.com/secureCodeBox/secureCodeBox/blob/e46688ecd2c71697a83e93c972e549847977679a/hooks/cascading-scans/hook/scan-helpers.ts#L138). Maybe something got lost during a refactor?

To reproduce, start any scan that triggers a cascading scan and check the logs of the cascading scan hook.

malexmave:
The cascading hook had a problem with logging the name of the cascaded scan (it was using a key that was undefined at the time it was used). This PR switches it to use the `metadata.generateName` key. The disadvantage of this approach is that it does not show the suffix that Kubernetes generates for the scan name - it will only log the combined scan name from parent and cascaded scan, but not the four-letter suffixes that are appended automatically. If there is a way to get these into the name as well, let me know, but I don't think there is (and, [judging from the Git history, this was also the behavior before it broke](https://github.com/secureCodeBox/secureCodeBox/blame/b38d8cffb2cd4b7b8139a564ce99981a31e5d53d/hooks/declarative-subsequent-scans/scan-helpers.ts#L80-L124)).

Closes #921.

### Checklist

### DIFF
--- a/hooks/cascading-scans/hook/scan-helpers.ts
+++ b/hooks/cascading-scans/hook/scan-helpers.ts
@@ -139,7 +139,7 @@ export async function startSubsequentSecureCodeBoxScan(scan: Scan) {
 
   try {
     // Submitting the Scan to the kubernetes api
-    await k8sApiCRD.createNamespacedCustomObject(
+    const createdScan = await k8sApiCRD.createNamespacedCustomObject(
       "execution.securecodebox.io",
       "v1",
       namespace,
@@ -147,6 +147,7 @@ export async function startSubsequentSecureCodeBoxScan(scan: Scan) {
       scan,
       "false"
     );
+    console.log(`-> Created scan ${createdScan.body["metadata"].name}`)
   } catch (error) {
     console.error(`Failed to start Scan ${scan.metadata.generateName}`);
     console.error(error);
--- a/hooks/cascading-scans/hook/scan-helpers.ts
+++ b/hooks/cascading-scans/hook/scan-helpers.ts
@@ -135,7 +135,7 @@ export function mergeInheritedSelector(parentSelector: LabelSelector = {}, ruleS
 }
 
 export async function startSubsequentSecureCodeBoxScan(scan: Scan) {
-  console.log(`Starting Scan ${scan.metadata.name}`);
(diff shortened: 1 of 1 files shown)


EndPositive:
@malexmave `createNamespacedCustomObject` returns the created object, so you should be able to retrieve the final name from it.

malexmave:
Thanks for the info - but it is only called after the first relevant logging statement, and the second log is only printed if the call failed, so I am not sure if that is a good approach 😅 - or am I misunderstanding something somewhere?

EndPositive:
Yeah it wouldn't work with the current logs, but you could do something like

```js

  console.log(`Creating Scan ${scan.metadata.generateName}...`);

  try {
    // Submitting the Scan to the kubernetes api
    const scan = await k8sApiCRD.createNamespacedCustomObject(
      "execution.securecodebox.io",
      "v1",
      namespace,
      "scans",
      scan,
      "false"
    );
    console.log(`-> Created scan ${scan.metadata.generateName}...`)
  } catch (error) {
    console.error(`Failed to create Scan ${scan.metadata.generateName}`);
    console.error(error);
  }
```

malexmave:
Ah, I see. Okay, I implemented this change, with a small change because the IDE did not believe me that the object would have the expected format, so I had to use the bracket syntax instead of the dot syntax. Thank you again for the proposal.

EndPositive (approved):
Looks good to me :smile:

J12934 (approved):

PR #936, opened 2022-01-19
ADR-0011: Version Numbers

### LINKED ISSUE #926: Establish and document what changes constitute a major/minor/patch version
SCB relies on many third-party scanners, some of which will perform major version upgrades occasionally (e.g., [gitleaks](https://github.com/secureCodeBox/secureCodeBox/pull/830), [sslyze](https://github.com/secureCodeBox/secureCodeBox/pull/844), ...). At the same time, we also make changes to the SCB codebase, which can also occasionally include major changes (updates to CRDs, ...).

At the moment, the policy for what consitutes a major release (SCB v2.X => v3.X) vs. a minor release (v3.5 => 3.6) vs. a patch release (v3.6.0 => 3.6.1) is unclear (wrt. included scanners) and inconsistently followed (wrt. changes to CRDs, for example). This issue tracks our process for agreeing on a new, explicit versioning scheme.

We discussed this issue at length today. From my recollection, this is what we came up with.

# General Points
The purpose of [(semantic) versioning](https://semver.org/) is that users can see directly "how worried they have to be" when installing an update. A single look at the old and new version number should tell them what kind of changes to expect, and if updating will require manual actions to keep the system running. We have always tried to follow semantic versioning principles, but have in some cases not done this correctly, and never really defined how it interacts with the updates released by external scanners. Thus, we now explicitly define what these terms mean in the context of the secureCodeBox.

# Major releases (v3.X => 4.0)
A *major* release should happen every time we change something that will cause failures on updates unless additional, manual actions are taken. This includes:
- Backwards-incompatible changes to scanners (e.g., requiring new parameters / changes to the scan YAML, as is the case for Gitleaks v8)
- Changes to CRDs (since they have to be installed manually)
- Breaking changes to the finding format of a scanner (i.e., changing the contents and meaning of an existing field, removing a field)
- ...

A major release does not mean that every single user of the system will have to make changes - it is enough if users using a specific scanner (gitleaks) will have to make changes, either to how they parameterize the scan or for working with the output format. The patchnotes should highlight exactly what part of the system contains a breaking change, and what actions are required to address it (either directly in the patch notes, or in a linked pull request / documentation page).

This policy implies that major releases will be a lot more common in the future, as especially the rule about changing CRDs was not followed in the last few months. However, it also implies that minor/patch releases should be installable with a lot more confidence as soon as this policy is in effect.

# Minor releases (v3.1.X => 3.2.0)
Following the principles of [semver](https://semver.org/), minor releases contain changes in functionality that are backwards-compatible. This includes:
- Updates to scanners, hooks or operators that add new functionality, but do not significantly alter the behavior of existing functionality
- Changes to the findings format that add new information without changing what is already there
- ...

A major release should, generally, be installable without reading the release notes (although we of course do not recommend this 😁), as it by definition does not require making manual changes after installation and does not significantly alter the functioning of whatever you have already configured.

# Patch releases (v3.1.0 => 3.1.1)
Patch releases only contain bug fixes that do not add new functionality, only repair what is already there. Importantly, a patch release can alter the behavior of the system, including the contents of findings, as long as what was previously being returned was incorrect. For example, if a finding listed a URL as `https:///google.com`, changing it to `https://google.com` (without the extra, incorrect slash) would be considered a patch, not a minor or major change.

-------

We are seeking input on this proposed policy, both from the rest of the team that wasn't present in the meeting today and from external contributors and users of the system like @EndPositive and @sanmai-NL. Please give your input in this issue. @Weltraumschaf @rseedorff, if I misremembered anything or you have more examples to add, please feel free.

Once we have agreed on a specific policy, the following changes will have to be made:
- [ ] Take a look at [UPGRADING.md](https://github.com/secureCodeBox/secureCodeBox/blob/main/UPGRADING.md), [SECURITY.md](https://github.com/secureCodeBox/secureCodeBox/blob/main/SECURITY.md), [README.md](https://github.com/secureCodeBox/secureCodeBox/blob/main/README.md). Make sure the information there matches the policy and is not outdated or redundant.
- [ ] Create a new section "Upgrading" in the web documentation that contains information on how to upgrade and what to expect, based on this policy.
- [ ] Rename our [GitHub project](https://github.com/orgs/secureCodeBox/projects/5) so it no longer references a specific major version. Migrate issues from the v4 project to the new main project and give them "breaking" labels if appropriate.
- [ ] Take a look at the release-drafter template and check if it requires changes based on this policy
- [ ] ...anything else?

malexmave:
This PR contains a new ADR on how we want to version the secureCodeBox. To quote the included ADR:

> Software version numbers should serve as an indicator of compatibility between different versions, both during operation and while updating to a newer release.
> At the moment, the _secureCodeBox_ is following the [Semantic Versioning (semver)](https://semver.org/) approach.
> According to semver, version numbers are of the format MAJOR.MINOR.PATCH, and the different places are incremented as follows:
> 
> > 1. MAJOR version when you make incompatible API changes,
> > 2. MINOR version when you add functionality in a backwards compatible manner, and
> > 3. PATCH version when you make backwards compatible bug fixes.
> 
> However, the architecture of _secureCodeBox_, with its operator and many individual scanners that can be installed and used separately, makes this seemingly simple versioning system more difficult.
> For example, we need to answer the following questions:
> 
> 1. Are breaking changes in the parameterization of a scanner (e.g. nmap, Gitleaks) a breaking change for the entire _secureCodeBox_ project?
> 2. Are changes to the output format of a single scanner a breaking change for the entire project?
> 3. In an environment where operator and scanners aren't necessarily using the same version number because the operator is controlled by a different team than the scanners (a setup that we want to support), how do we indicate compatibility between different versions of operator and scanner?
> 
> Depending on how these questions are answered, different versioning schemes are possible.

This PR is intended as a place to iterate on the ADR and discuss the options until a decision is reached. If you disagree with anything in the ADR, feel free to use the code review functionality to start a discussion.

I have made one fairly fundamental assumption that underlies the ADR: We know that some users of the SCB have a shared Kubernetes cluster with a single SCB operator and different namespaces where different teams can install their own scanners, set up rules, etc. - in these cases, the operator and scanners may not be upgraded in lockstep. **The ADR assumes that this is a correct and supported way of using the secureCodeBox.** If there is disagreement here, we should discuss that before discussing versioning schemes, because it makes a large difference as to which options are viable.

Closes #926.


EndPositive:
@malexmave, I think the idea of separating Operator and Scanner versioning is quite interesting. If there is a bugfix in the operator that does not influence the scanners, why should I update the scanner charts?

I think it would be nice to have a versioning like this:

| Component | Major | Minor | Patch |
| --- | --- | --- | --- |
| Operator |  Deleting/renaming CRD | Adding nullable CRD | Performance / Backwards-compatible bugfixes |
| Scanners / Hooks | Identical to Major version of Operator that it supports |   helm chart changes + backwards-compatible SDK changes | scanner version |

Example:
Operator v2.1.1
Nmap: v2.0-7.92-r2

v2 in nmap version indicates that it is compatible with any v2.x.x of the operator.

I think this scheme is fairly similar to option 3. My proposal is not that polished, but a solution like that sounds pretty good to me.

EndPositive:
Hi @J12934 , thanks for adding my option to the ADR. However, I think there is a small misunderstanding. My proposal had the scanners/hooks version number only include the major version of the operator, instead of the complete version.

Again example:
Operator v2.1.1
Nmap: v2.0-7.92-r2

The example from the ADR: 
operator: 3.42.0
nmap: 3.42.0+7.92-r2

See how in the ADR option 5, the full operator version is specified. I purposely include only the major version in the nmap version so that small changes in the operator don't require scanner versions to be updated. I think that is also why it is more similar to option 3 than to 2.

malexmave:
I like your proposal in general. I am not quite sure how it would handle a case where we make an improvement to a scanner without updating the version of the scanner itself though. Like saying "oh, actually, there is another piece of interesting data in the output of the scanner, let's quickly change the parser to also expose this in the findings". Since the embedded scanner remains on the same version, and the operator also remains the same, how would a version update like this be numbered, especially in an automated fashion? Similarly, what if we update the parser base image and need to re-generate all parser images - how would this reflected in the versioning scheme you propose, @EndPositive? The solution by @J12934 has the advantage that we have the entire version number to play with, which makes changes like this easy.

Or, put another way: Why is it bad that the scanner version needs to be updated for small changes to the operator (because this seems to be the basic premise of your proposal)? At the moment, we are releasing numbered versions for all parts of the system on every release, regardless of if they changed or not. This way, you can always just pull all version numbers up on upgrade, and if you see that parts of your stack are running on 3.3.1 and parts on 3.4, you don't need to wonder if this is because 2.4 did not make any changes to them or because you forgot to pull this component up to v3.4.

EndPositive:
> Since the embedded scanner remains on the same version, and the operator also remains the same, how would a version update like this be numbered, especially in an automated fashion?

That's what I reserved the minor version for. E.g.:

nmap: v2.0-7.92-r2 -> v2.1-7.92-r2

Updating the SDK would result in all scanners/hooks releasing with a minorversion update. That might be a bit confusing when multiple scanners are on the same SDK, but not on the same minor (e.g. because one has updated params).

> Why is it bad that the scanner version needs to be updated for small changes to the operator (because this seems to be the basic premise of your proposal)?

Like you mentioned earlier, it is very likely that third-party developers maintain their own scanners. I can't imagine they release a new scanner version for each release that is made here. Simply because there may not be a requirement to do so. How many times do we make a release for scanners that have 0 changes?

By modularization like this I hope you can also find that there is less maintenance in releasing new scb versions.

It is a common concept to modularize like this. ZAP addons also update independently from ZAP. Nuclei templates are also modular from nuclei engine. Differnece between these and my proposal was that  the core version is not mentioned in module version, only in changelog.

docs/adr/adr_0011.md
```diff
@@ -0,0 +1,167 @@
...
+> Any change that requires manual actions from at least one user of the SCB to keep existing workflows running after an update is considered a breaking change and requires a MAJOR release.
+
+This manual action may include making changes to scan definitions, or to systems that are ingesting data from the SCB findings of a scanner.
+It sees the entire _secureCodeBox_ as **one large piece of software with many components** that are all equally important to the overall compatibility, and where all components are (usually) updated in lockstep.
+As illustration, here are a few examples and what kind of release they would require:
+
+| Action                                                                             |Version |
+|------------------------------------------------------------------------------------|--------|
+| A scanner changes how it is parameterized                                          | Major  |
+| A scanner changes what data it returns                                             | Major  |
```
Weltraumschaf, line 45: Is this really major because this is abstracted by our parser layer. So in an ideal world we can compensate this change inside the parser.

docs/adr/adr_0011.md
```diff
@@ -0,0 +1,167 @@
...
+# ADR-0011: Version Numbers
+
+| <!-- -->       | <!-- --> |
+|----------------|----------|
+| **Status**:    | DRAFT |
+| **Date**:      | 2022-01-18 |
+| **Author(s)**: | Max Maass <max.maass@iteratec.com> |
+
+## Context
+Software version numbers should serve as an indicator of compatibility between different versions, both during operation and while updating to a newer release.
```
Weltraumschaf, line 16: IMHO it would make sense to summarize all our APIs where breaking changes may occur. E.g. the CRDs, the Finding Output Format, the Parser SDK etc. Based on that we can decide which of this API has impact to whom and how. E.g. an Parser SDK API change will not affect an operator of the SCB, but a maintainer of a scanner.

docs/adr/adr_0011.md
```diff
@@ -0,0 +1,167 @@
...
+1. **The proposal deviates from the SemVer standard** and is thus unexpected for people who do not know about it.
+
+### Option 4: Semantic Versioning, Separate Versioning For Components
+For completeness sake, we also include this option, in which each component is versioned separately, so that a breaking change in one component only changes the version number of that component.
+We previous discarded this idea because it would require complex documentation of which scanner versions work with which operator versions.
+We will thus not consider this proposal in greater detail here.
+
+### Option 5: SemVer With Major Version Indicating Operator Compatibility + Included Version of "Embedded Dependency"
+
+(Based on Suggestion by @EndPositive: https://github.com/secureCodeBox/secureCodeBox/pull/936#issuecomment-1016480293)
```
Weltraumschaf, line 117:
```suggestion
(Based on [Suggestion](https://github.com/secureCodeBox/secureCodeBox/pull/936#issuecomment-1016480293) by @EndPositive)
```

docs/adr/adr_0011.md
```diff
@@ -0,0 +1,167 @@
...
+This option expands on option 2, but adds an additional the version of embedded dependencies (primarily of scanner) in the build metadata section of the semver version (See: https://semver.org/#spec-item-10).
```
Weltraumschaf, line 118:
```suggestion
This option expands on option 2, but adds an additional version of embedded dependencies (primarily of scanner) in the build [metadata section of the semver](https://semver.org/#spec-item-10) version.
```

Weltraumschaf: This is new to me. Long ago that I've read the SemVer specs.

Weltraumschaf:
> By modularization like this I hope you can also find that there is less maintenance in releasing new scb versions.

We had modularization in SCBv1 and it was a hell. That's why we decided to maintain everything like a monolithic software w/ single version. Turns out it is not sufficient :-(

Complete other idea: Is it feasible to somehow implement a dry-run upgrade, so that I can check if soething will break? If this is possible we may not need any version.

docs/adr/adr_0011.md
malexmave: This was referring more to a case like gitleaks, where data is completely removed from the output, or the format is changed significantly (i.e., something that cannot be compensated through changes in the parser). I'll reword this part of the table.

malexmave: Is the new version clearer?

docs/adr/adr_0011.md
malexmave: I don't quite understand what exactly you propose to do here, so I have not made changes to address this comment. What would be the impact on the versioning from this differentiation? You want to know which kind of change is breaking to whom? For which purpose - being able to give more specific upgrade guidance in the release notes, or should it have a direct impact on the versioning itself?

Weltraumschaf: Usually in an architecture documentation you have something like a [component overview](https://docs.arc42.org/section-5/). It is the first level drill down of the whole system. In this component overview each component describes which APIs it provides.

If we would have something like that – not here in the ADR but as general architecture documentation – we'll have an overview to identify all the interfaces of components which we need to cover by this ADR. At the moment we are discussing possible breaking changes here and there, but IMHO we do not share common picture of:

* what components do we have
* what APIs do the use to interface each other
* which of these APIs must be considered for versioning

Weltraumschaf: This is wip now.

Weltraumschaf (dismissed):
LGTM but not sure which option we should choose.

SebieF:
When releasing breaking changes, it would probably be nice for users to be noticed upon installation/upgrading. This could happen not only by putting the breaking release note prominently at the top of the release notes, but also to have them in the *Helm notes* of the respective scanners (The process of doing that should be automated, of course..)

malexmave:
I have updated the ADR with our decision in the meeting we had today.

This decision is made with the knowledge that it may have consequences we are unaware of. We thus invite people who disagree with it to let us know by creating a discussion or opening an issue.

It results in the following new TODOs:

- [ ] #1023
- [ ] Explain the versioning scheme in the documentation, likely as part of the [forthcoming "upgrading" page](https://github.com/secureCodeBox/documentation/issues/232) 
- [ ] https://github.com/secureCodeBox/secureCodeBox/issues/1024

rseedorff (approved):

PR #944, opened 2022-01-26
Helm - Juice Shop - Custom environment variables and application config.yml

fbuchmeier-abi:
## Description

Added two new options to the Juice Shop Helm Chart:

- `customConfig` allows passing in a custom config file in yaml format
- `containerEnv` allows setting custom container environment variables (for the juice-shop container), e.g. to use a different config file

Quick question, which license header to I have to add to the new `configmap.yaml` and what implications does this have?

Thanks!

### Checklist

* [ x ] Test your changes as thoroughly as possible before you commit them. Preferably, automate your test by unit/integration tests.

### DIFF
--- a/demo-targets/juice-shop/values.yaml
+++ b/demo-targets/juice-shop/values.yaml
@@ -21,13 +21,13 @@ imagePullSecrets: []
 nameOverride: ""
 fullnameOverride: ""
 
-# add custom environment variables to the juice-shop container
+# -- add custom environment variables to the juice-shop container
 # ref: https://pwning.owasp-juice.shop/part1/customization.html
 containerEnv: []
   # - name: NODE_ENV
   #   value: bodgeit
 
-# When set, passes in the custom configuration specfied below and sets NODE_ENV accordingly
+# -- When set, passes in the custom configuration specified below and sets NODE_ENV accordingly
 # ref: https://pwning.owasp-juice.shop/part1/customization.html
 customConfig:
   # NOTE: Ensure that all required keys are passed, otherwise startup will fail
--- a/demo-targets/juice-shop/values.yaml
+++ b/demo-targets/juice-shop/values.yaml
@@ -21,6 +21,25 @@ imagePullSecrets: []
 nameOverride: ""
 fullnameOverride: ""
 
+# add custom environment variables to the juice-shop container
+# ref: https://pwning.owasp-juice.shop/part1/customization.html
+containerEnv: []
+  # - name: NODE_ENV
+  #   value: bodgeit
+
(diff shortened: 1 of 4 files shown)


J12934:
Hi @fbuchmeier-abi thanks for the contribution looks good 👍

Not a lawyer so can't give you any legal guidance but my understanding is:

- The License header should match the copyright field from the License file: https://github.com/secureCodeBox/secureCodeBox/blob/main/LICENSES/Apache-2.0.txt#L61
- The license header is more of a precaution in case somebody copies / takes a single file out of the context of the repo
- The sign-off flag on the commit gives the repo owner the (copy?)right to potentially use the code in a context outside the current Apache 2.0 license, e.g. release a new version of the SCB under MIT license without having to collect the signature of all previous contributors.

I'd like to change the copyright holder in the future to some "entity" which isn't just "iteratec", like "secureCodeBox core team" but not sure what the entails legally so can't promise anything 🤞

demo-targets/juice-shop/values.yaml
```diff
@@ -21,6 +21,25 @@ imagePullSecrets: []
 nameOverride: ""
 fullnameOverride: ""
 
+# add custom environment variables to the juice-shop container
```
J12934, line 24:
```suggestion
# -- add custom environment variables to the juice-shop container
```

demo-targets/juice-shop/values.yaml
```diff
@@ -21,6 +21,25 @@ imagePullSecrets: []
...
+# ref: https://pwning.owasp-juice.shop/part1/customization.html
+containerEnv: []
+  # - name: NODE_ENV
+  #   value: bodgeit
+
+# When set, passes in the custom configuration specfied below and sets NODE_ENV accordingly
```
J12934, line 30:
```suggestion
# -- When set, passes in the custom configuration specified below and sets NODE_ENV accordingly
```

J12934 (commented):
Some small suggestions to properly include the descriptions / comments in side the helm docs page shown on artificthub.io.

(These are automatically generated using: https://github.com/norwoodj/helm-docs)

fbuchmeier-abi:
@J12934 thank you for the suggestions and the explanation. I need to check with my coworker first if the license header is OK for is. I'll update you once this is done.

Regards,
Florian.

fbuchmeier-abi:
@J12934 I've updated the PR with the license header as requested: https://github.com/secureCodeBox/secureCodeBox/pull/944/commits/78149d2c78eb2482f437cdd3d64402743f3a56eb

Anything else I can do?

Regards,
Florian.

J12934:
Perfect, thank you 🚀

J12934 (approved):

PR #946, opened 2022-01-27
Fix Email Reciever Field in Notification Hook

J12934:
## Description

The Email Notifier of the Notification Hook currently behaves differently as described in the Docs.
The docs indicate that `channel.endPoint` is supposed to be email recipient for the notification but actually has to be the name of a `env` var which then contains the email address.
This is a feature was added for the Slack & Microsoft Teams Notifier as these use webhook urls containing sensitive tokens.

I've updated the implementation to behave like the documentation.

### Checklist

### DIFF
--- a/hooks/notification/.helm-docs.gotmpl
+++ b/hooks/notification/.helm-docs.gotmpl
@@ -252,7 +252,11 @@ notificationChannels:
     endPoint: "someone@somewhere.xyz"
 env:
   - name: SMTP_CONFIG
-    value: "smtp://user:pass@smtp.domain.tld/"
+    # you can create the secret via: kubectl create secret generic email-credentials-token --from-literal="smtp-config=smtp://user:pass@smtp.domain.tld/"
+    valueFrom:
+      secretKeyRef:
+        name: email-credentials
+        key: smtp-config
 ```
 
 To provide a custom `from` field for your email you can specify `EMAIL_FROM` under env.
@@ -261,7 +265,10 @@ For example:
 ```
 env:
   - name: SMTP_CONFIG
-    value: "smtp://user:pass@smtp.domain.tld/"
+    valueFrom:
+      secretKeyRef:
+        name: email-credentials
+        key: smtp-config
   - name: EMAIL_FROM
     value: secureCodeBox
 ```
--- a/hooks/notification/.helm-docs.gotmpl
+++ b/hooks/notification/.helm-docs.gotmpl
@@ -252,7 +252,7 @@ notificationChannels:
     endPoint: "someone@somewhere.xyz"
 env:
   - name: SMTP_CONFIG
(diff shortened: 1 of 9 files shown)


hooks/notification/hook/Notifiers/EMailNotifier.test.ts
```diff
@@ -80,7 +82,40 @@ test("Should Send Mail", async () => {
...
+low: 2<br>
+informational: 1<br>
+
+<br>
+<strong>Findings Category Overview:</strong><br>
+A Client Error response code was returned by the server: 1<br>
+Information Disclosure - Sensitive Information in URL: 1<br>
+Strict-Transport-Security Header Not Set: 1<br>
+`,
+    subject: "New nmap security scan resulsts are available!",
```
malexmave, line 103: Seems like there is a typo somewhere in the code 😁 (proposed edit here just to highlight the typo, the source will be somewhere else).
```suggestion
    subject: "New nmap security scan results are available!",
```

malexmave (commented):
Thanks for fixing this. I have a small comment on a typo. Also, since you mentioned the documentation: It is currently a bit inconsistent. At the top, it says "The general configuration of a notification looks something like this" and then shows an example that loads the env from a secret. However, for the email hook, it then loads the env directly from the file, with the username and password for the email account visible. Maybe we can take this opportunity to also fix that small part of the docs? (Reworking the rest of the docs for the hook is tracked in https://github.com/secureCodeBox/documentation/issues/182 and does not need to be addressed in this PR.)

hooks/notification/hook/Notifiers/EMailNotifier.test.ts
J12934: Ah haha yes, I thought i fixed it but this was probably in a commits I've later reverted 😅

J12934: Should be resolved now

hooks/notification/.helm-docs.gotmpl
```diff
@@ -252,7 +252,11 @@ notificationChannels:
     endPoint: "someone@somewhere.xyz"
 env:
   - name: SMTP_CONFIG
-    value: "smtp://user:pass@smtp.domain.tld/"
+    # you can create the secret via: kubectl create secret generic email-credentials-token --from-literal="smtp-config=smtp://user:pass@smtp.domain.tld/"
+    valueFrom:
+      secretKeyRef:
+        name: email-credentials
```
malexmave, line 258: This seems inconsistent: `email-credentials-token` (in the kubectl create) vs. `email-credentials` (in the secretKeyRef) - can you double-check this?

malexmave (changes requested):

malexmave (approved):

PR #948, opened 2022-02-01
[SCB-Bot] Upgraded trivy from 0.22.0 to 0.24.2

secureCodeBoxBot:
This is an automated Pull Request by the SCB-Bot. It upgrades trivy from 0.22.0 to 0.24.2
### Release changes : [0.23.0](https://github.com/aquasecurity/trivy/releases/v0.23.0), [0.24.0](https://github.com/aquasecurity/trivy/releases/v0.24.0), [0.24.1](https://github.com/aquasecurity/trivy/releases/v0.24.1), [0.24.2](https://github.com/aquasecurity/trivy/releases/v0.24.2)


malexmave:
This release [contains a breaking change for parameterization](https://github.com/aquasecurity/trivy/discussions/1515): `trivy [image name]` is no longer supported, only `trivy image [image name]` is. We will have to double-check the docs, tests, etc., and add a warning in the release.

It also has a change that is labelled as breaking [about the SARIF templates](https://github.com/aquasecurity/trivy/pull/1437) (where I don't understand what exactly is happening or breaking), and about the source of the rule download (where the [rules are now downloaded from the GitHub Container Registry instead of GitHub releases](https://github.com/aquasecurity/trivy-db/pull/159), which means that there should be fewer issues with rate limiting 🥳). This will also have to be checked in the documentation.

malexmave:
Also, [a new feature where you can specify dockerhub credentials directly in the Helm values](https://github.com/aquasecurity/trivy/pull/1569). @J12934 is this relevant for our deployment, or are we using a different helm chart anyway?

J12934:
> Also, [a new feature where you can specify dockerhub credentials directly in the Helm values](https://github.com/aquasecurity/trivy/pull/1569). @J12934 is this relevant for our deployment, or are we using a different helm chart anyway?

We don't use that chart directly.  But interesting to see, didn't even know that trivy had a official helm chart.

malexmave:
Okay, I thought this would be a quick update, but it turns out that it is actually complicated. Documenting the issues here for when we get back to it.

Trivy can be launched with different commands. `trivy image` used to be the default, and could be shortened to a plain `trivy [image name]`. This has been removed with this update, so you always have to write `trivy image`. So far, so easy.

The problem is that the trivy argument parser really likes to have its arguments in order. So, `trivy image --format json bkimminich/juice-shop:v10.2.0` works, but `trivy --format json image bkimminich/juice-shop:v10.2.0` will fail. Since we need to provide a bunch of parameters (`--no-progress --format json --output /home/securecodebox/trivy-results.json`) for the scanner to work as intended, this raises a problem: We have to explicitly provide the "image" command before adding these parameters, but since we add the parameters in the ScanType we would also have to provide the `image` command there. However, if we provide the `image` command in the scan type, users can no longer use non-`image` subcommands of trivy, like `trivy fs` for file system scans. But if we do not provide the `image` command and extra parameters, we would have to instruct the user to provide the parameters themselves, which is annoying as it exposes the "plumbing" of secureCodeBox to the user.

I see four options for this:
- Only support `image` scans. I don't like this, as the other types are definitely useful.
- Force users to provide the parameters themselves on every scan. Annoying, will lead to breakage.
- Change the ScanType handling in the operator to give an extra field for appending instead of prepending arguments to the final command. Requires changes to the operator and would thus be a larger change. I am not sure if this would completely resolve the issue, as I do not know if all trivy subcommands actually support the parameters.
- Provide our own docker container with a wrapper script that appends the relevant parameters. Same problem with being unsure if the parameters will be supported for all subcommands, but at least it does not require changes to the operator.

Opinions are welcome.

J12934:
I think the easiest (and maybe best) thing to do here is to split up the trivy scan type into multiple ones.
Similar to how we do it for ZAP: (https://github.com/secureCodeBox/secureCodeBox/blob/main/scanners/zap/templates/zap-scan-type.yaml)

Then we still have one trivy chart, but it installs 2 (or more?) scanTypes `trivy-image` & `trivy-fs`.

rseedorff:
👍🏻 Let's split the scanTypes here as we already did for the ZAP scantype.

J12934:
I've updated the pull request to use the multiple scan types.
Didn't yet have the time to test it properly, especially the trivy-filesystem scanType.

rseedorff (approved):

PR #954, opened 2022-02-03
Improve Makefiles for generating other documentation

Weltraumschaf:


Weltraumschaf:
All targets are fixed and work on macOS.

malexmave (approved):
Works great, tested it for scanners, hooks, demo-targets, operator and auto-discovery.

Issue #953, opened 2022-02-03
SCB Bot doc generation deviates from CI doc generation


malexmave:
## 🐞 Bug report
The SCB-Bot is keeping our dependencies up to date, and also runs a documentation updater as part of that process to regenerate the helm docs. It seems like the process it uses deviates from that of the CI helm-docs update, because the two seem to be fighting - the bot [changes the docs in a specific way](https://github.com/secureCodeBox/secureCodeBox/pull/952/commits/6951029770744a0cad873834842bdb18b7ca41b7) (that seems incorrect to me), the other [reverts the change](8faa2b85252abcd6d0b04469099c805344d1168d). Both commits come from https://github.com/secureCodeBox/secureCodeBox/pull/952/.

### Expected behavior
I would expect the doc generation process of the bot to match that of the CI process, so we don't have unnecessary edit wars between our different automations 😁 .

Weltraumschaf:
Seems to be resolved.

PR #955, opened 2022-02-04
Upgrade SCB-Bot helm-docs version to 1.6.0

malexmave:
We updated helm-docs for the helm-docs CI job to 1.6.0, but did not do the same for the SCB-Bot, which may be the root cause of https://github.com/secureCodeBox/secureCodeBox/issues/953 (working theory, will have to be validated with the next PR opened by the Bot after this is merged).

### DIFF
--- a/.github/workflows/scb-bot.yaml
+++ b/.github/workflows/scb-bot.yaml
@@ -102,9 +102,9 @@ jobs:
           mkdir helm-docs
           cd helm-docs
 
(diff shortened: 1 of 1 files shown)


Weltraumschaf (approved):
We should extract this download and verify shell snippet into a shell script in `bin/`, so that we do not have this code duplication. Which obviously resulted in a bug^^

PR #1002, opened 2022-02-25
ZapAdvanced Enforced Spider Timeouts

J12934:
## Description

This PR enforces the max duration feature of ZAP if ZAP is unable to do it by its own.
This prevents Scenarios where scans are stuck endlessly in the spidering phase.

To make this easier I have refactored the spider waiting logic so that the waiting is handled by the parent class shared by the traditional and ajax spider. (b039dd95d1d77af05cfdcff9601ba1e7d9174d88)
This was required to implement the maxDuration logic for both without duplicating the code.

This is unfortunately pretty hard to test as the cases where the Spider hangs are hard to reproduce.
I've tested the enforcement logic locally by overwriting the maxDuration set for ZAP to a much higher value so that the ZapAdvanced maxDuration enforcement mechanism could kick in.

### Checklist

### DIFF
--- a/scanners/zap-advanced/scanner/zapclient/spider/zap_abstract_spider.py
+++ b/scanners/zap-advanced/scanner/zapclient/spider/zap_abstract_spider.py
@@ -8,6 +8,7 @@
 
 import collections
 import logging
+import time
 
 from abc import abstractmethod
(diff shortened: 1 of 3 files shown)


rseedorff (approved):
Like your approach here 👍🏻

PR #1009, opened 2022-03-01
Updated ScheduledScan's operator tests to increase code coverage

Ilyesbdlala:
## Description

- Current tests are converted to the BDD [ginkgo style](https://onsi.github.io/ginkgo/).
- Added Operator test to check that findings get updated; Findings summary should be updated for the scheduled scan when a child scan is triggered.

### Checklist

### DIFF
--- a/operator/controllers/execution/scheduledscan_controller_test.go
+++ b/operator/controllers/execution/scheduledscan_controller_test.go
@@ -5,7 +5,6 @@ package controllers
 
 import (
 	"context"
-	"reflect"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


operator/controllers/execution/scheduledscan_controller_test.go
```diff
@@ -1,63 +1,106 @@
...
+
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, types.NamespacedName{Name: "test-scan", Namespace: namespace}, &scheduledScan)
+				if errors.IsNotFound(err) {
+					panic("ScheduledScan should be present for this check!")
+				}
+				return (scheduledScan.Status.Findings.Count == 42 &&
+					scheduledScan.Status.Findings.FindingSeverities == executionv1.FindingSeverities{High: 42} &&
+					reflect.DeepEqual(scheduledScan.Status.Findings.FindingCategories, map[string]uint64{"Open Port": 42}))
+			}, timeout, interval).Should(BeTrue())
```
J12934, line 103: I'd try to use the Ginko Assertions as much as possible as these will print out a more clear diff when something goes wrong. Something like this:

(Suggestion can't be applied directly as this also removes the need for the reflect import)

```suggestion
			Eventually(func() bool {
				err := k8sClient.Get(ctx, types.NamespacedName{Name: "test-scan", Namespace: namespace}, &scheduledScan)
				if errors.IsNotFound(err) {
					panic("ScheduledScan should be present for this check!")
				}
				return scheduledScan.Status.Findings.Count != 0
			}, timeout, interval).Should(BeTrue())

			Expect(scheduledScan.Status.Findings.Count).Should(Equal(uint64(42)))
			Expect(scheduledScan.Status.Findings.FindingSeverities).Should(Equal(executionv1.FindingSeverities{High: 42}))
			Expect(scheduledScan.Status.Findings.FindingCategories).Should(Equal(map[string]uint64{"Open Port": 42}))
```

J12934 (commented):
Nice Work 👍

One small suggestion, which i think would make the assertions a bit easier to understand

J12934 (approved):

PR #1014, opened 2022-03-04
[SCB-Bot] Upgraded nuclei from v2.6.2 to v2.6.5

secureCodeBoxBot:
This is an automated Pull Request by the SCB-Bot. It upgrades nuclei from v2.6.2 to v2.6.3
### Release changes : [here](https://github.com/projectdiscovery/nuclei/releases/v2.6.3)


rseedorff:
This patch upgrade [adds a new severity "unkown"](https://github.com/projectdiscovery/nuclei/pull/1659): 
"Possible severity values: info, low, medium, high, critical, **unknown**" maybe this change needs to be reflected in our parser somehow 🤔

J12934:
I'd personally map unkown to informational. 🤷‍♂️

Ilyesbdlala:
> I'd personally map unkown to informational. man_shrugging

I don't think mapping the findings to informational is a good idea. We usually reserve this to findings that don't result in vulnerabilities (like the service version in NMAP). Quoting the nuclei issue concerning this:

> For the cases where the impact of the issue is not proved/confirmed as a result of the finding and can be decided after further inspection, at the same time using info is also not accurate for those templates.

Personally, I see the findings of type "unknown" deserving more attention than an Informational finding, and our mapping should reflect that. Maybe setting it as "LOW"  would be me better idea.

malexmave:
I just looked through a couple of scanners that I am familiar with: 
- In the `gitleaks` scanner, we [map unknown severities to MEDIUM](https://github.com/secureCodeBox/secureCodeBox/blob/main/scanners/gitleaks/parser/parser.js#L17-L23).
- In `semgrep` [the default is informational](https://github.com/secureCodeBox/secureCodeBox/blob/main/scanners/semgrep/parser/parser.js#L26).

So, seems like there isn't a globally recognized default (and by @Ilyesbdlala argument, we should probably also switch semgrep to a default of LOW). I agree with Ilyes that for nuclei, LOW seems a better default than INFORMATIONAL.

rseedorff (approved):

rseedorff:
Thx for your feedback! Switched the mapping to a LOW severity in case nuclei claims a unknown.

PR #1022, opened 2022-03-08
Removed dependency on Dockerhub for Test-Scan scantypes in integration tests

### LINKED ISSUE #964: Remove dependency on Dockerhub for Test-Scan scantypes in integration tests
Currently, all integration tests that use the test-scan scantype rely on pushing and pulling images from DockerHub. As part of our refactoring of the CI, we want to get rid of this dependency. Proposed change:

- Have a single job that runs the integration tests for the test-scan scantype itself
- Build (w/o running the tests) the scantype in every other CI job that requires it (instead of pulling from DockerHub)

Ilyesbdlala:
closes #964 

Signed-off-by: Ilyes Ben Dlala <ilyes.bendlala@iteratec.com>

### Checklist


.github/workflows/ci.yaml
```diff
@@ -769,32 +698,50 @@ jobs:
...
+        working-directory: ./parser-sdk/nodejs/
+        run: make docker-build-sdk
+
+      - name: Export Docker Images
+        working-directory: ./parser-sdk/nodejs/
+        run: make docker-export
+
+      - name: Kind import
+        working-directory: ./parser-sdk/nodejs/
+        run: make kind-import
```
malexmave, line 718: I think that these two steps are unnecessary - the SDK image does not need to be present in the cluster, it is only needed as a base image to build the parsers later. For the parsers, these two steps are correct, but here they don't hurt, but also don't do anything productive, afaik.

malexmave (commented):
I'm not the right person to review this for correctness, as I don't know enough about how the CI works right now, but here's at least one comment that came up while I was going through the diff.

.github/workflows/ci.yaml
RamiSouai: You are correct. The steps are now gone.

Weltraumschaf (approved):
LGTM

PR #1028, opened 2022-03-09
Inherit Makefile Integration Tests command for hooks 

RamiSouai:
## Description

The integration-test command for hooks is now the same for all hooks, and uses npm run tests (With the exception of Persistance-DefectDojo which is in Java). 

Now each hook can decide for itself what the `npm run test` command does in it's package.json file. For example, notification and cascading-scans hooks, which are in Typescript, add an extra build step.

### Checklist


hooks.mk
```diff
@@ -42,4 +42,4 @@ deploy:
 integration-tests:
 	@echo ".: 🩺 Starting integration test in kind namespace 'integration-tests'."
 	kubectl -n integration-tests delete scans --all
-	cd ../../tests/integration/ && npm ci &&	npx --yes --package jest@$(JEST_VERSION) jest --verbose --ci --colors --coverage --passWithNoTests ${hook-prefix}s/$(name).test.js
+	cd .. && npm ci && cd $(name)/${hook-prefix} && npm run test --package jest@$(JEST_VERSION)
```
Weltraumschaf, line 45: Is there a system behind the naming convention, if a var is lower-kebab-case or upper-snake-case? Usually in Makefiles is upper-snale-case like in Shell scripts. I recommend one style, unless there is a really good and documented reason for this exception.

hooks/cascading-scans/Makefile
```diff
@@ -54,15 +56,5 @@ deploy:
...
 		--set="hook.image.repository=docker.io/$(IMG_NS)/$(hook-prefix)-$(name)" \
 		--set="hook.image.tag=$(IMG_TAG)"
-
-integration-tests:
-	@echo ".: 🩺 Starting integration test in kind namespace 'cascading-scans'."
-	kubectl -n integration-tests delete scans --all
-	cd ../../tests/integration/ && npm ci &&	npx --yes --package jest@$(JEST_VERSION) jest --verbose --ci --colors --coverage --passWithNoTests scanner/cascade-nmap-ncrack.test.js
-
+		
 deploy-test-deps-2: deploy-test-dep-unsafe-https deploy-test-dep-nmap deploy-test-dep-sslyze
```
Weltraumschaf, line 60: In general make uses the target name as file name, e.g. you can have a target `*.ts` which will be invoked for all TypeScript files. You use this to call the compiler in the target. Typical targets from C are `*.c`, *.o` etc. All targets which are not file names are so called [phony targets](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html) and must be tagged with `.PHONY`. Unless we do this a target may be invoked ifsoeone adds a file which accidentally match the target name.

Please add `.PHONY` to all targets.

Weltraumschaf (changes requested):
LGTM. Only .PHONY tags we should add to at least all touched Makefiles.

hooks/cascading-scans/Makefile
```diff
@@ -8,10 +8,14 @@
 include_guard = set
 hook = cascading-scans
 
+test_files = [hook.test.js, kube.test.js]
+
 include ../../hooks.mk
 
+.PHONY: test-2
 test-2: | clean-integration-tests unit-tests docker-build docker-export kind-import deploy deploy-test-deps-2 integration-tests-2
```
Weltraumschaf, line 16: What does the `|` here? Never seen this before and I'm too stupid to google it...

rseedorff: * https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html 
* https://stackoverflow.com/questions/58039810/makefiles-what-is-an-order-only-prerequisite

Weltraumschaf: Second question: What does the 2 mean in the target name? We should avoid magic numbers in names.

PR #1033, opened 2022-03-15
Improves the integration-tests makefile process for scanners

Ilyesbdlala:
This PR removes the scanners` dependency on integration tests located in tests/integration

* The tests are moved to their corresponding scanner folder (instead of /tests/integration/)
* The integration-tests makefile target is only overwritten when necessary:
   * by git-repo-scanner: Integration-tests are disabled to avoid GitHub API rate limit
   * by semgrep: to allow the use of a config-map
   * by test-scan: since it has not yet been migrated to the makefile integration-tests template
* The `scanners.mk` is changed to make use of the integration-tests located in the $(scanner-name)/integration-tests folder  

As a side effect of the transition to the use of makefile. We've had to have a two copies of helpers.js. One in scanners/ folder and the other in tests/integration.  The one in tests/integration can be deleted after the migration of the hooks integration tests is complete. (This is why the codeclimate check is unhappy)

### Checklist


scanners/helpers.js
```diff
@@ -0,0 +1,284 @@
+// SPDX-FileCopyrightText: the secureCodeBox authors
```
Weltraumschaf, line 1: Lot of going on here... Is this file copied from somewhere else or written from scratch? In the latter case we should also have at least some basic unit tests for the more complex functions (with for-loops and such) to prevent simple errors.

Ilyesbdlala: It's copied over as mentioned in the PR description:
> As a side effect of the transition to the use of makefile. We've had to have a two copies of helpers.js. One in scanners/ folder and the other in tests/integration. The one in tests/integration can be deleted after the migration of the hooks integration tests is complete. (This is why the codeclimate check is unhappy)

But yes unit tests should be implemented.

Weltraumschaf: ok

Weltraumschaf (approved):

PR #1034, opened 2022-03-15
Fixes semgrep's integration test error

Ilyesbdlala:
replaces the expected result of integration test  of semgrep to high severity
This is done because the vulnerabilities found have increased in severity according to semgrep scanner.

Signed-off-by: Ilyes Ben Dlala <ilyes.bendlala@iteratec.com>

### Checklist

### DIFF
--- a/scanners/semgrep/integration-tests/semgrep.test.js
+++ b/scanners/semgrep/integration-tests/semgrep.test.js
@@ -35,7 +35,7 @@ test(
       "security": 3,
     });
     expect(severities).toEqual({
-      medium: 3,
+      high: 3,
     });
   },
   3 * 60 * 1000


malexmave (approved):
Oh, the semgrep integration tests use a rule from the semgrep registry instead of one defined inline? That's probably a bad practice as well on my part when I implemented that. Thanks for fixing the severity. In the medium term, we probably also want to replace it with a local rule that does not interact with the registry, right? (Doesn't have to be done in this PR, just wondering if I should open an issue for that)

Ilyesbdlala:
> Oh, the semgrep integration tests use a rule from the semgrep registry instead of one defined inline? That's probably a bad practice as well on my part when I implemented that. Thanks for fixing the severity. In the medium term, we probably also want to replace it with a local rule that does not interact with the registry, right? (Doesn't have to be done in this PR, just wondering if I should open an issue for that)

I think so, but would we be then no longer testing the "registry usage" of the semgrep scanner? If a problem occurs in that process we would not notice AFAIK.

malexmave:
True, but that would be a bug in semgrep and not in SCB, right?

Although, to be fair, I am reasonably sure that we have other tests that are interacting with other online registries as well, so it's probably not a big issue to have this small interaction with outside systems.

PR #1050, opened 2022-03-18
Adding kind cluster name variable to common.mk makefile config (closes #1037)

### LINKED ISSUE #1037: Introduce a makefile config for local kind cluster name
A minor comment here (I am not entirely sure if this belongs directly to this issue): When I have a `kind` cluster with a different name (for example: scb), kind related make targets will fail, e.g.:
```bash
common-kind-import:
	@echo ".: 💾 Importing the image archive '$(module)-$(name).tar' to local kind cluster."
	kind load image-archive ./$(module)-$(name).tar  #fix: --name scb
```
I think it is the only affected make-target.

_Originally posted by @SebieF in https://github.com/secureCodeBox/secureCodeBox/issues/1027#issuecomment-1065044803_

SebieF:
## Description

The PR, once applied, makes sure that the kind command in common.mk also works with local clusters that have a different name than "kind". `kubectl config-current-context` gives the current kind cluster, but its output looks like "kind-scb" for a kind cluster named "scb", so you have to cut off the prefix with the cut command.

(closes #1037)

### Checklist

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -43,9 +43,12 @@ K := $(foreach exec,$(EXECUTABLES),\
 #
 # JEST_VERSION  		Defines the jest version used for executing the tests. Defaults to latest
 #
+# KIND_CLUSTER_NAME:	Defines the name of the kind cluster (created by kind create cluster --name cluster-name)
+#
 # Examples:
 # 	make all IMG_TAG=main
 # 	make deploy IMG_TAG=$(git rev-parse --short HEAD)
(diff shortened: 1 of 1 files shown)


rseedorff (changes requested):

Weltraumschaf (approved):

Weltraumschaf:
Don't know y GH thinks that the change @rseedorff requested is not done yet. For me it looks like done.

Issue #1019, opened 2022-03-07
common.mk does not use python3 (correctly)


SebieF:
### Describe the bug

The file [common.mk](https://github.com/secureCodeBox/secureCodeBox/blob/main/common.mk) used by our Makefiles uses "python" instead of "python3". "python" is the old command for Python 2. Creating an alias with `alias python='python3'` did not work for me in the makefile.
Furthermore, pip and pytest are used without a python command, which results in an error for me.

### Steps To Reproduce
`make unit-tests` in scanners/gir-repo-scanner

### System (please complete the following information):
Ubuntu 20.04 LTS

### Potential fix
Line 29:
Use **python3**
`EXECUTABLES = make docker kind git node npm npx kubectl helm yq java python3`  

Line 79-85:
Add python3:
```bash
install-deps-py:
	@echo ".: ⚙️ Installing all $(module) specific python dependencies."
	python3 -m pip install --upgrade pip setuptools wheel pytest
	cd ./$(module)/ && python3 -m pip install -r requirements.txt

unit-test-py: install-deps-py
	cd ./$(module)/ && python3 -m pytest --ignore-glob='*_local.py' --ignore=tests/docker
```

I think this version will work for most users. *The reason why I did not create a PR immediately is, however, that I don't know how this will behave on MacOS or Windows and it could be potentially breaking there.*

Weltraumschaf:
As a work around until we fix this you can use `python-is-python3` package, if you're on Debian/Ubuntu based machines.

malexmave:
Can confirm that this behavior also occurs on OS X when `python` is not available, but `python3` is.

Weltraumschaf:
This is a more complex problem bc python may be at various location depending on the OS or distro (e.g. /usr/bin/python, /usr/bin/python3, /usr/local/bin/python, /usr/local/nin/python3 etc.). 

As a workaround you can use:
* on macOS [PyEnv installieren](https://londonappdeveloper.com/installing-python-on-macos-using-pyenv/)
* on Linux the package `python-is-python3`

In general the Makefile must determine the right path to Python. Common pattern in Makefiles is:

```Makefile
PYTHON = ...

$PYTHON ....
```

So instead of using the command directly in the targets you use the variable `$PYTHON` which holds the absolute path to the wanted executable.

How to determine the value of `$PYTHON` the right way needs further investigation. Never done that before. We should look into other Makefiles which use Python.

Weltraumschaf:
[...]

PR #1053, opened 2022-03-21
Minor makefile improvements for the test-all target

rseedorff:
### Checklist


.python-version
```diff
@@ -0,0 +1 @@
+3.9.10
```
Weltraumschaf, line 1: You should really add some words in the commit message why this is done. I've no clue how this the problem works around.

Weltraumschaf (approved):

.python-version
rseedorff: Good point 🤔  this is just a first workaround for those using pyenv on macos (should be fixed for all envirnments in #1019)

PR #1071, opened 2022-03-29
Make common.mk use python3 (Closes #1019)

### LINKED ISSUE #1019: common.mk does not use python3 (correctly)

SebieF:
## Description
On some systems, the command `python` is either non-existing or actually calling `python3`. This PR, once applied, makes sure, that the correct command is selected and saved in the new `$(PYTHON)` variable and that Python version >= 3.0 is actually used (which is necessary for the makefile).

Also re-enables deactivated unit tests for git-repo-scanner.

To test if the PR works on your system, run the following commands:
```bash
cd scanners/git-repo-scanner
make unit-tests
# Should finish with all tests passed
```

Closes #1019.

### Checklist
Works and tested for the following setups:
- [x] Linux (Ubuntu)
- [x] Linux with Anaconda
- [ ] MacOs 
- [ ] PyEnv

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -25,8 +25,23 @@ ifeq ($(name),)
   $(error name ENV is not set)
 endif
 
+PYTHON = $(shell which python3)
+ifeq ($(PYTHON),)
+  PYTHON = $(shell which python)
+  ifeq ($(PYTHON),)
+  	$(error "PYTHON=$(PYTHON) not found in $(PATH)")
+  endif
+endif
+PYTHON_VERSION_MIN=3.0
+PYTHON_VERSION=$(shell $(PYTHON) -c \
+'import sys; print(float(str(sys.version_info[0]) + "." + str(sys.version_info[1])))')
+PYTHON_VERSION_OK=$(shell $(PYTHON) -c 'print(int($(PYTHON_VERSION) >= $(PYTHON_VERSION_MIN)))' )
+ifeq ($(PYTHON_VERSION_OK), 0) # True == 1
+   $(error "Need python version >= $(PYTHON_VERSION_MIN) (current: $(PYTHON_VERSION))")
+endif
+
 # Thx to https://stackoverflow.com/questions/5618615/check-if-a-program-exists-from-a-makefile
-EXECUTABLES = make docker kind git node npm npx kubectl helm yq java python
+EXECUTABLES = make docker kind git node npm npx kubectl helm yq java $(PYTHON)
 K := $(foreach exec,$(EXECUTABLES),\
         $(if $(shell which $(exec)),some string,$(error "ERROR: The prerequisites are not met to execute this makefile! No '$(exec)' found in your PATH")))
 
@@ -84,11 +99,11 @@ unit-test-js: install-deps-js
 
 install-deps-py:
 	@echo ".: ⚙️ Installing all $(module) specific python dependencies."
-	python -m pip install --upgrade pip setuptools wheel pytest
-	cd ./$(module)/ && pip install -r requirements.txt
+	$(PYTHON) -m pip install --upgrade pip setuptools wheel pytest
+	cd ./$(module)/ && $(PYTHON) -m pip install -r requirements.txt
 
 unit-test-py: install-deps-py
-	cd ./$(module)/ && pytest --ignore-glob='*_local.py' --ignore=tests/docker
+	cd ./$(module)/ && $(PYTHON) -m pytest --ignore-glob='*_local.py' --ignore=tests/docker
 
 unit-test-java:
 	cd ./$(module)/ && ./gradlew test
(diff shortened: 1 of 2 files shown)


malexmave:
Can confirm that the updated Makefile works on MacOS.

common.mk
```diff
@@ -25,8 +25,23 @@ ifeq ($(name),)
...
+'import sys; print(float(str(sys.version_info[0]) + "." + str(sys.version_info[1])))')
+PYTHON_VERSION_OK=$(shell $(PYTHON) -c 'print(int($(PYTHON_VERSION) >= $(PYTHON_VERSION_MIN)))' )
+ifeq ($(PYTHON_VERSION_OK), 0) # True == 1
+   $(error "Need python version >= $(PYTHON_VERSION_MIN) (current: $(PYTHON_VERSION))")
+endif
+
 # Thx to https://stackoverflow.com/questions/5618615/check-if-a-program-exists-from-a-makefile
-EXECUTABLES = make docker kind git node npm npx kubectl helm yq java python
+EXECUTABLES = make docker kind git node npm npx kubectl helm yq java $(PYTHON)
 K := $(foreach exec,$(EXECUTABLES),\
```
Weltraumschaf, line 45: What is `K`?

Weltraumschaf (commented):
Left a question.

Weltraumschaf:
> Would this solve the issue with the zap-advanced nightly build unit-test step? https://github.com/secureCodeBox/secureCodeBox/runs/5713889414 The step "make unit-tests" works locally, but fails on the CI. I have a feeling it has to do with the discrepancy you mentioned.
> 
> Would this solve the issue with the zap-advanced nightly build unit-test step? https://github.com/secureCodeBox/secureCodeBox/runs/5713889414
The step "make unit-tests" works locally, but fails on the CI. I have a feeling it has to do with the discrepancy you mentioned.

Not sure. The log output of the action is quite useless about the cause.

common.mk
SebieF: `$(info $$K is [${K}])` gives `$K is [ some string  some string  some string  some string  some string  some string  some string  some string  some string  some string  some string  some string]`

It is not a change of the PR, but the [link](https://stackoverflow.com/questions/5618615/check-if-a-program-exists-from-a-makefile) mentioned in the comment above states that "K" is necessary as an arbitrary variable name in order to have the right hand side executed immediately.

SebieF:
> Would this solve the issue with the zap-advanced nightly build unit-test step? https://github.com/secureCodeBox/secureCodeBox/runs/5713889414 The step "make unit-tests" works locally, but fails on the CI. I have a feeling it has to do with the discrepancy you mentioned.
> 
> P.S: The Amass issue has been resolved in the main branch. Upping the Amass appVersion or merging main will solve your pipeline issue.

I'm afraid it won't solve the issue.. In the environment variables of the run it says:
` env:
    GO_VERSION: 1.17
    PYTHON_VERSION: 3.9
    NODE_VERSION: 16
    NPM_VERSION: 7
    DOCKER_NAMESPACE: ***`

So Python availability doesn't seem to be an issue there.

common.mk
Weltraumschaf: Maybe you should choose a better name for this variable.

SebieF: It is not part of the PR, should I change it anyway?

Weltraumschaf: Yes, that's the pathfinder rule of clean code: leave the place cleaner than you found it. This rule helps to improve the whole code base step by step over time.

SebieF: @Weltraumschaf Alright I refactored the name :) If possible, could you check if it is running with PyEnv? I think you are using it..

Weltraumschaf: works:

```
» pyenv version
3.9.10 (set by /Users/sst/src/work/scb/secureCodeBox/.python-version)
...
tests/git_repo_scanner_test.py ........                                                                                                                                                                                              [100%]

============================================================================================================ 8 passed in 0.62s =============================================================================================================
```

Weltraumschaf (approved):

Weltraumschaf (changes requested):
There are unsigned commits.

SebieF:
> There are unsigned commits.

Only merge commits are unsigned, is it possible/needed to sign them (DCO does not complain..)? I am afraid to use something like `git rebase --signoff HEAD~2` because of all of the merged commits that do not need to be signed..

Weltraumschaf:
> > There are unsigned commits.
> 
> Only merge commits are unsigned, is it possible/needed to sign them (DCO does not complain..)? I am afraid to use something like `git rebase --signoff HEAD~2` because of all of the merged commits that do not need to be signed..

All of your commits are not signed with a digital signature (either S/MIME or PGP). I do not mean sign-off.

SebieF:
Closed due to issues with signing.

Weltraumschaf:
Upsi, looks like a failed rebase. Maybe this helps for future tasks: https://learngitbranching.js.org/

PR #1102, opened 2022-04-12
Improved the local makefile workflow

### LINKED ISSUE #756: Makefile-based Scanner integration test flow is very cumbersome to run locally
## Acceptance Criteria:
- In each *Scanner* directory (with existing integrationtests) implement an valid `integration-test` target: `make integration-test`
- Review & Update existing documentation regarding "integration-tests": https://docs.securecodebox.io/docs/contributing/integrating-a-scanner/makefile
- There is no longer a general integration-test folder `/secureCodeBox/tests/integration/`-> all hook related integrationtests are moved 

- [ ] Check AMASS
- [ ] Check angularjs-csti-scanner
- [ ] Check cmseek
- [ ] Check git-repo-scanner
- [ ] Check gitleaks
- [ ] Check kube-hunter
- [ ] ...

### Long read

At the moment, running integration tests locally is fairly painful. The goal is that a single `make integration-tests` in a scanner / hook directory should Just Work, but it very much does not at the moment.

In the worst case (modifications to the operator, no previous compiled state), in my experience the following is the minimum set of commands to run integration tests for a scanner (it may be that there is a simpler way, but I have not found it):

```bash
# If you also need to reset your cluster:
# kind delete cluster
# kind create cluster

# Prepare operator
cd operator
make docker-build
make docker-export
make kind-import
rm operator.tar lurker.tar  # clean up
make helm-deploy

# Prepare parser and hook SDK as dependency for the parser/hook
# In case of a scanner
cd ../parser-sdk/nodejs/
make docker-build-sdk
# In case of a hook
cd ../../hook-sdk/nodejs/
make docker-build-sdk

# Run the actual integration tests, with the example of semgrep
cd ../../scanners/semgrep/
make docker-build
make docker-export
make kind-import
rm parser-semgrep.tar scanner-semgrep.tar  # clean up
make deploy
make deploy-test-deps
# Ideally, the following is the only line I would need for all of this, maybe excluding the operator
make integration-tests
```

Some of this can probably be solved by settings dependencies inside the Makefile, but this may then require changes to the CI pipeline as well to avoid duplicate work being done there.

In an ideal world, `make integration-tests` in a scanner / hook directory would:
- Check if the cluster is present, and give a warning if it is not
- Validate the state of the cluster (is the operator running?) and get it into a good state if it isn't already
- Compile and load all dependencies for the docker build (i.e., hook / parser sdk)
- Compile and load the current scanner into the cluster
- Start the integration tests

I am marking this as both a bug and an enhancement because I was told that theoretically, this should already be the case with the current makefiles and it may be that the functionality was lost over time.

Ilyesbdlala:
## Description

closes #756 

Allows the following : 

- Running `make test` in a hook or scanner folder would run a clean install and run the corresponding unit-tests/integrations-tests
- Running `make test-scanner target=<SCANNER-NAME>` in the Project root directory, runs `make test` in the corresponding Scanner folder
- Running `make test-hook target=<SCANNER-NAME>` in the Project root directory, runs `make test` in the corresponding hook folder

### Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,42 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-PROJECT_DIR					= $(shell pwd)
-BIN_DIR						= $(PROJECT_DIR)/bin
-SCANNERS_DIR				= $(PROJECT_DIR)/scanners
-HOOKS_DIR					= $(PROJECT_DIR)/hooks
-DEMO_TARGETS_DIR			= $(PROJECT_DIR)/demo-targets
-OPERATOR_DIR				= $(PROJECT_DIR)/operator
-PARSER_SDK_DIR				= $(PROJECT_DIR)/parser-sdk/nodejs
-HOOK_SDK_DIR				= $(PROJECT_DIR)/hook-sdk/nodejs
-AUTO_DISCOVERY_DIR			= $(PROJECT_DIR)/auto-discovery
-HELM_DOCS_DIR				= $(PROJECT_DIR)/.helm-docs
-TEMPLATES_DIR				= $(PROJECT_DIR)/.templates
-
-SCANNERS_CHART_LIST			:= $(sort $(wildcard $(SCANNERS_DIR)/*/Chart.yaml))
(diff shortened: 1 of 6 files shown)


RamiSouai (approved):
LGTM

PR #1107, opened 2022-04-12
Feature/authenticated generic webhook

Miker91:
The PR is related to the **generic-webhook** ScanCompletionHook.

This small feature introduces the possibility to post a generic webhook including credentials. 

This can be helpful in CI/CD integration when the destination system requires authentication.

Credentials can be stored as Kubernetes secret.

### DIFF
--- a/hooks/generic-webhook/values.yaml
+++ b/hooks/generic-webhook/values.yaml
@@ -31,3 +31,12 @@ hook:
 
   # hook.tolerations -- Optional tolerations settings that control how the hook job is scheduled (see: https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/)
   tolerations: []
+
+  # hook.authentication -- Optional basic authentication credentials 
+  authentication:
+    # -- Link a pre-existing generic secret with `username` and `password` key / value pairs
+    userSecret: generic-webhook-credentials
+    # -- Name of the username key in the `userSecret` secret. Use this if you already have a secret with different key / value pairs
+    usernameKey: username
+    # -- Name of the password key in the `userSecret` secret. Use this if you already have a secret with different key / value pairs
+    passwordKey: password 
--- a/hooks/generic-webhook/values.yaml
+++ b/hooks/generic-webhook/values.yaml
@@ -34,9 +34,10 @@ hook:
 
   # hook.authentication -- Optional basic authentication credentials 
   authentication:
-    # -- Link a pre-existing generic secret with `usernameKey` and `passwordKey` key / value pairs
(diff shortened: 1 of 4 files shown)


Ilyesbdlala:
Thank you for your contribution !
In our project, we follow [DCO](https://developercertificate.org/). Each commit, as mentioned [here](https://github.com/secureCodeBox/secureCodeBox/blob/main/CONTRIBUTORS.md), has to be signed off.
Running the steps found [here](https://github.com/secureCodeBox/secureCodeBox/pull/1107/checks?check_run_id=5993384904) should do the trick.
The CI pipeline should hopefully run on your next git push.

Miker91:
@Ilyesbdlala Thanks for the advice. Hopefully it is ok now.

hooks/generic-webhook/values.yaml
```diff
@@ -31,3 +31,12 @@ hook:
...
   tolerations: []
+
+  # hook.authentication -- Optional basic authentication credentials 
+  authentication:
+    # -- Link a pre-existing generic secret with `usernameKey` and `passwordKey` key / value pairs
+    userSecret: generic-webhook-credentials
+    # -- Name of the username key in the `userSecret` secret. Use this if you already have a secret with different key / value pairs
+    usernameKey: username
+    # -- Name of the password key in the `userSecret` secret. Use this if you already have a secret with different key / value pairs
+    passwordKey: password
```
J12934, line 42: I think it might be worth it to add an additional layer here, to clearly mark that the username and password are used for basic auth and also to allow adding other authentication mechanisms (like the auth header) later.

```suggestion
  # hook.authentication -- Optional basic authentication credentials 
  authentication:
    basic:
      # -- Link a pre-existing generic secret with `usernameKey` and `passwordKey` key / value pairs
      userSecret: generic-webhook-credentials
      # -- Name of the username key in the `userSecret` secret. Use this if you already have a secret with different key / value pairs
      usernameKey: username
      # -- Name of the password key in the `userSecret` secret. Use this if you already have a secret with different key / value pairs
      passwordKey: password
```

Note: suggestion can't be applied directly as it also requires a change to the helm template where the vars are used :)

Miker91: Good point. More methods is yet to come:)

rseedorff:
Hi @Miker91 
thx for your contribution 🙌 and welcome to our community :)
Can you please add yourself to the [CONTRIBUTORS.md list](https://github.com/secureCodeBox/secureCodeBox/blob/main/CONTRIBUTORS.md) as part of our DCO?

J12934 (approved):

PR #1133, opened 2022-04-25
Make common.mk use python3 (new) (Closes #1019) 

### LINKED ISSUE #1019: common.mk does not use python3 (correctly)

SebieF:
## Description
On some systems, the command `python` is either non-existing or actually calling `python3`. This PR, once applied, makes sure, that the correct command is selected and saved in the new `$(PYTHON)` variable and that Python version >= 3.0 is actually used (which is necessary for the makefile).

Also re-enables deactivated unit tests for git-repo-scanner.

To test if the PR works on your system, run the following commands:
```bash
cd scanners/git-repo-scanner
make unit-tests
# Should finish with all tests passed
```

Closes #1019.

### Checklist
Works and tested for the following setups:
- [x] Linux (Ubuntu)
- [x] Linux with Anaconda
- [x] MacOs 
- [x] PyEnv 

*This is a new version of the old pull-request #1071 which had a failed rebase and unsigned commits.*

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -25,8 +25,23 @@ ifeq ($(name),)
   $(error name ENV is not set)
 endif
 
+PYTHON = $(shell which python3)
+ifeq ($(PYTHON),)
+  PYTHON = $(shell which python)
+  ifeq ($(PYTHON),)
+  	$(error "PYTHON=$(PYTHON) not found in $(PATH)")
+  endif
+endif
+PYTHON_VERSION_MIN=3.0
+PYTHON_VERSION=$(shell $(PYTHON) -c \
+'import sys; print(float(str(sys.version_info[0]) + "." + str(sys.version_info[1])))')
+PYTHON_VERSION_OK=$(shell $(PYTHON) -c 'print(int($(PYTHON_VERSION) >= $(PYTHON_VERSION_MIN)))' )
+ifeq ($(PYTHON_VERSION_OK), 0) # True == 1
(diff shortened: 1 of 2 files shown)


Weltraumschaf (approved):

Weltraumschaf:
Some checks were not successfull.

SebieF:
> Some checks were not successfull.

@Weltraumschaf Fixed now :)

PR #1141, opened 2022-04-28
Update OWASP status from Incubator to Lab in more READMEs

malexmave:
While going through the docs, I realized that the shields in the scanner READMEs still list the project as an OWASP Incubator project. I changed the template (following the way it was done in https://github.com/secureCodeBox/secureCodeBox/commit/ba821f733b85eaea80315fafddf7945ef82b8663) and regenerated all the READMEs using `make docs`.

This resulted in four README files not being updated and still containing the Incubator status:
- lurker/docs/README.DockerHub-Core.md
- scanners/gitleaks/docs/README.DockerHub-Hook.md
- scanners/gitleaks/docs/README.DockerHub-Scanner.md
- scanners/semgrep/docs/README.DockerHub-Scanner.md

To me it seems like the second was generated incorrectly and can be removed (gitleaks is not a hook). Since gitleaks and semgrep don't have a dedicated scanner image at the moment, these files are likely also outdated and can be removed. However, I don't quite understand why the lurker does not receive updates. It does not have its own `.helm-docs.gotmpl` file with chart-specific information, but it does *contain* chart-specific info (how to pull the lurker image, the info that it is intended to be used with the operator and not standalone, ...

So, my questions are:
1. Should I just delete files 2-4, or am I missing something?
2. How is file 1 updated?

Once that is resolved, this PR can be merged to correct the docs.


SebieF:
I agree that 2-4 should be deleted. 

The reasons why 1 (lurker docs) is not updated are that:
1. There is no command for it in [make docs](https://github.com/secureCodeBox/secureCodeBox/blob/ab386f8ba3950e7a6093fcb27e6f944f8bdcae4d/Makefile#L69). Thus there are also no commands in the generate-helm-docs script etc.
2. There is no Chart.yaml file in the `lurker` directory. Even if `helm-docs` is called in the lurker directory, it won't update the `README.DockerHub-Core.md` file.

I am not sure, however, if this issue should be resolved by creating a Chart.yaml file in the lurker directory or making it possible to update the `README.DockerHub-Core.md` even if there is no Chart.yaml file in this very directory.

rseedorff (approved):
LGTM

Weltraumschaf:
We talk about this file: `lurker/docs/README.DockerHub-Core.md`

PR #1151, opened 2022-05-03
Use venv when installing deps with pip

### LINKED ISSUE #1150: Use venv for python deps
The make target `install-deps-py` does not use a venv and thereby will install required pip packages systemwide. This is really bad because then python packages installed by the packagemanger and pip are mixed and there is no way to distinguish them (at least on arch linux).
'install-deps-py`should use a venv in order to not install random pyhton packages on  system level.

the-simmon:
This PR closes #1150.
The make target `install-deps-py` and `unit-test-py` will now use a virtual environment (venv) instead of installing packages system wide.
This can be tested by running `make unit-tests` in the zap-advaned folder.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -103,11 +103,11 @@ $(PYTHON_VENV_NAME):
 
 install-deps-py: | $(PYTHON_VENV_NAME)
 	@echo ".: ⚙️ Installing all $(module) specific python dependencies."
-	./venv/bin/pip install --upgrade pip setuptools wheel pytest
-	./venv/bin/pip install -r $(module)/requirements.txt
+	./$(PYTHON_VENV_NAME)/bin/pip install --upgrade pip setuptools wheel pytest
+	./$(PYTHON_VENV_NAME)/bin/pip install -r $(module)/requirements.txt
 
 unit-test-py: install-deps-py
-	cd $(module)/ && ../venv/bin/python -m pytest --ignore-glob='*_local.py' --ignore=tests/docker
+	cd $(module)/ && ../$(PYTHON_VENV_NAME)/bin/python -m pytest --ignore-glob='*_local.py' --ignore=tests/docker
 
 unit-test-java:
 	cd $(module)/ && ./gradlew test
--- a/common.mk
+++ b/common.mk
@@ -98,11 +98,13 @@ unit-test-js: install-deps-js
 
 install-deps-py:
 	@echo ".: ⚙️ Installing all $(module) specific python dependencies."
-	$(PYTHON) -m pip install --upgrade pip setuptools wheel pytest
-	cd $(module)/ && $(PYTHON) -m pip install -r requirements.txt
+	@echo "Creating venv to avoid installing deps system wide"
+	$(PYTHON) -m venv venv
+	./venv/bin/pip install --upgrade pip setuptools wheel pytest
+	./venv/bin/pip install -r $(module)/requirements.txt
 
 unit-test-py: install-deps-py
-	cd $(module)/ && $(PYTHON) -m pytest --ignore-glob='*_local.py' --ignore=tests/docker
+	cd $(module)/ && ../venv/bin/python -m pytest --ignore-glob='*_local.py' --ignore=tests/docker
 
 unit-test-java:
 	cd $(module)/ && ./gradlew test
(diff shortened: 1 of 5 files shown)


common.mk
```diff
@@ -98,11 +98,13 @@ unit-test-js: install-deps-js
...
-	$(PYTHON) -m pip install --upgrade pip setuptools wheel pytest
-	cd $(module)/ && $(PYTHON) -m pip install -r requirements.txt
+	@echo "Creating venv to avoid installing deps system wide"
+	$(PYTHON) -m venv venv
+	./venv/bin/pip install --upgrade pip setuptools wheel pytest
+	./venv/bin/pip install -r $(module)/requirements.txt
 
 unit-test-py: install-deps-py
-	cd $(module)/ && $(PYTHON) -m pytest --ignore-glob='*_local.py' --ignore=tests/docker
+	cd $(module)/ && ../venv/bin/python -m pytest --ignore-glob='*_local.py' --ignore=tests/docker
```
Weltraumschaf, line 107: Here you should use `$(PYTHON)` and `$(PIP) instead of direct paths into the venv. The directory name `venv` may change. you can chose it at will (e.g. `python -m venv foo`. So `$(PYTHON)` and `$(PIP)` should point into the venv bin dir.

Also it should be considered to move this completely out of the Makefile. The idea here is to allow choice of installing system wide or use a venv: So we can provide a make target to create the venv and the user must activate it themself. From now on the python iterpreter is changed to the venv and make need not bother it. Also you can autoactivate/-deactivate e.g. with dirnv (`.envrc`):

```
# Activate Python's virtue environment
source venv/bin/activate
```

scanners/git-repo-scanner/Makefile
```diff
@@ -21,3 +21,23 @@ integration-tests:
 	#cd ../../tests/integration/ && npm ci
 	#cd ../../scanners/${scanner}
 	#npx --yes --package jest@$(JEST_VERSION) jest --verbose --ci --colors --coverage --passWithNoTests ${scanner}/integration-tests
+
```
Weltraumschaf, line 24: What has this to do with pip and venv? 🤔

scanners/zap-advanced/Makefile
```diff
@@ -30,6 +30,7 @@ kind-import-parser:
 	cd $(SCANNERS_DIR)/zap/ && $(MAKE) -s kind-import-parser
 
 deploy-with-scanner:
+	rm -rf venv # delete venv. Otherwise it would be deployed with zap
```
Weltraumschaf, line 33: This means that one need to reininstall the whole venv all the time which is time consuming. Is it possible to ignore the directory? I wonder if there is something like `.dockerignore`, maybe `.helmignore`?

Unless you should use a variable for the dir `venv` because as mentioned above the directory name can be chosen freely.

Weltraumschaf (changes requested):

common.mk
the-simmon: 1. `/bin/sh` does not support the `source` command. We need to change the Makefile shell to a different shell that supports source (the obvious choice would be bash, but I dont know if bash is a requirement of the SCB project already. What happens on systems without bash?). By calling the venv manullay we can still use `/bin/sh`.
Some distros might symlink `/bin/sh` to a different shell, but for me on Arch linux it does not support source.
2. Currently we give the user the choice to use a venv, but we dont inform the user about this. So running the seemingly harmless make target `unit-tests` installs python deps systemwide, which has the real possibility of bricking the OS. Additionally it is almost impossible to properly remove those installed pip packages. (This might be only true for certain linux distros, but for me on Arch it is horrible to install pip packages globally).
I would rather just a create a venv without asking the user then installing pip packages systemwide without asking. In my opinion there is no usecase for systemwide installation of pip packages. Not for a random unit test that people might run without looking at the makefile beforehand.

Weltraumschaf: I'm not convinced and the longer I think about it I come to the opinion that it is no good idea to do it this way. One problem you mentioned is /bin/sh used in make. This problem even goes deeper: Each line in a make target is executed in its own subshell. So activating the venv in one line has no effect for the following lines. But for workingf with venv the activation is crucial. The whole concept of venv relies on the fact that you activate the venv before doing things. In the approach above there is no activation. Just simply pointing to a pip command somewhere in the venv direcotry. This may work at the moment, but this is not realiable because venv implies its activation to work correctly.

So IMHO we should leave creating the venv out of the make target. Instead we should docuemnt the benefits of using venv (I know the pain of killing systems Python from macOS 😬). And we couldprovide a make target to create the venv as a short hand for someone never did this before.

IMHO this has the benefit that each user can decide to use venv, pyenv, pipenv or whatever.

PR #1199, opened 2022-05-30
Bugfixing Defaults for Notification Hook and ZAP Advanced Config templates

rseedorff:
### Checklist


hooks/notification/templates/configmap.yaml
```diff
@@ -2,9 +2,15 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+---
+{{- if not (empty .Values.notificationChannels) }}
```
J12934, line 6: This looks like this would cause more problems, as the ScanCompletionHook always tries to mount this config map.
Only creating it when it is set, will cause the hook pod to be in an infinite pending state. (Unless we set it to an optional mount)

J12934 (approved):

Issue #1200, opened 2022-05-31
Consistency: labels and annotations of scanner-, parser- and hook pods


malexmave:
At the moment, labels and annotations of parser- and scanner pods are fairly inconsistent.

Scanners:
- Labels:
  - `app.kubernetes.io/managed-by=securecodebox` (but not consistently - on one cluster, this is used (scans automatically generated by autodiscovery), on another cluster, it is not (scans set up manually)).
- Annotations:
  - `securecodebox.io/job-type: scanner` 

Parsers:
- Labels:
  - None
- Annotations:
  - `auto-discovery.securecodebox.io/ignore: true` 

Hooks:
- Labels:
  - None
- Annotations:
  - `auto-discovery.securecodebox.io/ignore: true`

Ideally:
- All scanners, hooks and parsers should have a label that states that they are managed by secureCodeBox
- All scanners, hooks and parsers should have the `auto-discovery.securecodebox.io/ignore: true` annotation
- All scanners, hooks and parsers should have a `securecodebox.io/job-type` annotation describing their type

This would allow more easily finding SCB-managed pods in kubectl commands using the `--field-selector` switch (at the moment, there is no single things that you can filter on to find all pods that are managed by the SCB).

May be related to https://github.com/secureCodeBox/secureCodeBox/issues/1194.

PR #1210, opened 2022-06-07
Make Labels and Annotation consistent

the-simmon:
This PR fixes https://github.com/secureCodeBox/secureCodeBox/issues/1200

### DIFF
--- a/operator/controllers/execution/scans/parse_reconciler.go
+++ b/operator/controllers/execution/scans/parse_reconciler.go
@@ -103,9 +103,13 @@ func (r *ScanReconciler) startParser(scan *executionv1.Scan) error {
 			BackoffLimit:            &backOffLimit,
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
+						"app.kubernetes.io/managed-by=securecodebox": "securecodebox",
+					},
 					Annotations: map[string]string{
 						"auto-discovery.securecodebox.io/ignore": "true",
 						"sidecar.istio.io/inject":                "false",
+						"securecodebox.io/job-type":              "parser",
 					},
 				},
 				Spec: corev1.PodSpec{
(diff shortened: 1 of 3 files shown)


operator/controllers/execution/scans/parse_reconciler.go
```diff
@@ -103,9 +103,13 @@ func (r *ScanReconciler) startParser(scan *executionv1.Scan) error {
 			BackoffLimit:            &backOffLimit,
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
+						"app.kubernetes.io/managed-by=securecodebox": "securecodebox",
+					},
```
malexmave, line 108: Is this pattern correct? `managed-by=securecodebox: securecodebox` seems redundant.

malexmave:
Thanks for your work! One request: Would it also be possible to add `securecodebox.io/job-type: scanner | parser | hook | operator | ...` labels in all relevant places? Currently, this seems to be handled as an annotation, and not for everything in SCB (see #1200). Making this a label would allow us to filter pods based on it using a podSelector, which I don't think is possible for annotations. A use case for this filtering would be in clusters using least-privilege networking with [Network Policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) - here, it may be desireable to permit egress to the Internet only for hooks, but not scanners and parsers (for example).

Although I am not sure if removing the annotation and turning it into a label would be considered a breaking change. Probably yes?

Unrelated: Do we have any idea why all the CI runs are failing (timing out) here? In some cases, it looks like they are generating good results, but the results are not pulled in correctly?

J12934:
> Although I am not sure if removing the annotation and turning it into a label would be considered a breaking change. Probably yes?

@malexmave 

Doesn't seem to be breaking (anymore not sure about before?) as this only adds new labels and doesn't remove anything.

@the-simmon are all remarks from Max addressed, code wise this looks good to me 👍

J12934:
@the-simmon very nice, thank you 👍

can you create a ticket for the annotations to remove and add them for the v4 board?

Issue #1193, opened 2022-05-25
Rename container autodiscovery scans


the-simmon:
Currently scheduled scans generated by the container autodiscovery are named `scan-image_name-at-image_hash`. The resulting scan will be called `scan-scan-image_name-at-image_hash`. To avoid the "scan-scan" the scheduled scans from the container autodiscovery should be renamed. This is a braking change, as the container autodiscovery will not "recognize" the old scans anymore. It will create new scans according to the new naming scheme. The old scheduled scans must be deleted manually.

PR #1243, opened 2022-06-23
Add comments describing scan name generation in ContainerAutodiscovery

the-simmon:
I just added some comments explaining how scan names are generated in the container autodiscovery.

### DIFF
--- a/auto-discovery/kubernetes/controllers/container_scan_controller.go
+++ b/auto-discovery/kubernetes/controllers/container_scan_controller.go
@@ -143,13 +143,17 @@ func cleanupImageID(imageID string, log logr.Logger) string {
 }
 
 func getScanName(imageID string) string {
+	//function builds string like: scan-_appName_-at-_imageID_HASH_ eg: scan-nginx-at-0123456789
 	baseScanName := "scan-"
+
+	//define appname cutoff limit to 20 chars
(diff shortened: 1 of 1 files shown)


auto-discovery/kubernetes/controllers/container_scan_controller.go
```diff
@@ -143,13 +143,17 @@ func cleanupImageID(imageID string, log logr.Logger) string {
 }
 
 func getScanName(imageID string) string {
+	//function builds string like: scan-_appName_-at-_imageID_HASH_ eg: scan-nginx-at-0123456789
 	baseScanName := "scan-"
```
SebieF, line 147: Thinking about it, this results in some "weird" scan names like: "scan-scan-nmap-xxxyz". Might be worth to change this in another PR to something like "auto-scan-..." to clarify the relation to the autodiscovery.

SebieF (approved):

auto-discovery/kubernetes/controllers/container_scan_controller.go
the-simmon: Sadly this would be a breaking change, so we have to wait for the next braking release. See https://github.com/secureCodeBox/secureCodeBox/issues/1193

Issue #1083, opened 2022-04-01
Error when using Wpscan scantype


RealAlphaMan:
Hello everyone, please help me with Wpscan scantype. I've setup wpscan and .yaml file.
But when I apply this and view the scan, its state is Errored, Error Description:   Failed to run the Scan Container, check k8s Job and its logs for more details. I checked log, it said "error: timed out waiting for the condition". Describe jobs: "BackoffLimitExceeded: Job has reached the specified backoff limit". 
Even if I added api-token parameter of Wpscan, this error also happened. I've try increased the backoffLimit but it's not better. I think the Scan Container inside pods has some issues. What happened with it?
```
apiVersion: "execution.securecodebox.io/v1"
kind: Scan
metadata:
  name: "wpscan-old-wordpress-internal"
spec:
  scanType: "wpscan"
  parameters:
    - "--url"
    - "http://old-wordpress/"
    - "-e"
    - "vp"
    - "--api-token"
     - "<MY_REGISTERED_TOKEN>"
```

Here is the logs from container Wpscans:
```
Command:
      wpscan
      -v
      -o
      /home/securecodebox/wpscan-results.json
      -f
      json
      --url
      http://old-wordpress/
      -e
      vp
      --api-token
      <MY_REGISTERED_TOKEN>
    State:          Terminated
      Reason:       Error
      Exit Code:    4 // Sometime I get 5
      Started:      Fri, 01 Apr 2022 17:35:11 +0700
      Finished:     Fri, 01 Apr 2022 17:35:13 +0700
    Last State:     Terminated
      Reason:       Error 
      Exit Code:    4 // Sometime I get 5
      Started:      Fri, 01 Apr 2022 17:35:09 +0700
      Finished:     Fri, 01 Apr 2022 17:35:11 +0700
```

J12934:
Hi 👋

Thanks for the report 👍

Have you installed the `old-workpres` demo application onto your k8s cluster before running the scan?

E.g. `helm install old-wordpress securecodebox/old-wordpress --version 3.9.0`

If so can you post the container logs of the pod? 
The outpout of `kubectl logs <pod_name> --all-containers`?
Note: depending on the kubernetes cluster kubernetes might clean up failed pods relativly quickly so you may have to be quick do do it.

the-simmon:
I cant recreate the issue. Here is what I did on a fresh kubernetes cluster using KIND:

[...]

PR #1260, opened 2022-07-05
Add Wpscan wrapper

the-simmon:
Solves https://github.com/secureCodeBox/secureCodeBox/issues/1083

Adds a wrapper to wpscan because wpscan does return non 0 values when it find vulnerability's.

### DIFF
--- a/scanners/wpscan/scanner/wrapper.sh
+++ b/scanners/wpscan/scanner/wrapper.sh
@@ -1,0 +1,7 @@
+# SPDX-FileCopyrightText: 2021 iteratec GmbH
+#
+# SPDX-License-Identifier: Apache-2.0
+
+wpscan $@
+echo "wpscan exited with $?"
+exit 0
--- a/scanners/wpscan/scanner/wrapper.sh
+++ b/scanners/wpscan/scanner/wrapper.sh
@@ -4,4 +4,7 @@
 
 wpscan $@
 echo "wpscan exited with $?"
+
+# wpscan returns a non zero exit code when it finds vulnerabilitys
+# see https://github.com/wpscanteam/CMSScanner/blob/master/lib/cms_scanner/exit_code.rb
 exit 0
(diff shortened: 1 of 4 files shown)


scanners/wpscan/scanner/wrapper.sh
```diff
@@ -0,0 +1,18 @@
...
+# wpscan returns a non zero exit code when it finds vulnerabilitys
+# exit codes 1, 2 ,3 and 4 are errors, 5 means vulnerabilitys were found
+# see https://github.com/wpscanteam/CMSScanner/blob/master/lib/cms_scanner/exit_code.rb
+fake_exit_code=0
+if [[ $wpscan_exit -eq 1 ]] || [[ $wpscan_exit -eq 2 ]] || [[ $wpscan_exit -eq 3 ]] || [[ $wpscan_exit -eq 4 ]]
+then
+    fake_exit_code=$wpscan_exit
+fi
+
+exit $fake_exit_code
```
SebieF, line 18:
```suggestion
if [[ $wpscan_exit -eq 5 ]]
then
    exit 0
fi

exit $wpscan_exit # No vulnerabilities found or an error has occurred
```

Might be a bit easier to read :)

SebieF (commented):
Only one small suggestion, rest looks good to me :)

J12934:
Hey @SebieF 

Well also need to add wpscan to the list of scanners to build a custom image for in the release build pipeline: https://github.com/secureCodeBox/secureCodeBox/blob/a6eae054bd4b2c927deaa77e4a196bc790cb8f4c/.github/workflows/release-build.yaml#L362

scanners/wpscan/values.yaml
```diff
@@ -31,7 +31,7 @@ parser:
 scanner:
   image:
     # scanner.image.repository -- Container Image to run the scan
-    repository: wpscanteam/wpscan
+    repository: docker.io/securecodebox/scanner-wpsan
```
rseedorff, line 34:
```suggestion
    repository: docker.io/securecodebox/scanner-wpscan
```
There seems to be a typo 🤔

rseedorff (changes requested):
woscan-scan-type.yaml:

`name: "wpscan{{ .Values.scanner.nameAppend | default ""}}"`

There are to many `"` chars in one string 🤔

J12934 (approved):

PR #1270, opened 2022-07-13
ADR-0012: Initial Proposal for Cluster Wide Custom Resources (e.g. ClusterScanType)

J12934:
This PR contain a Proposal for new variants for some of our CustomResourceDefinitions which could be used on a cluster level and then be reused across namespaces.

### DIFF
--- a/docs/adr/adr_0012.md
+++ b/docs/adr/adr_0012.md
@@ -237,7 +237,11 @@ The following (not so) edge cases should be considered:
 
 ### RBAC Concept
 
-**TODO:** Add considerations for how this will all interact with the RBAC system, how it is managed who is able to schedule scans, control targetNamespaces, etc.
+- The RBAC System / guidelines for existing CRDs stays the exact same.
+- The newly added ClusterScoped CRDs should be read only to normal cluster users by default.
+- Only cluster admins should be allowed to create, edit, patch and delete the cluster scoped CRDs
+
+This ensures that the security model remains intact as only cluster level users should be able to interact with resources shared across all resources and makes sure that it can't be used by attackers to escalate their privileges into other namespaces.
 
 ## Decision
 We use Proposal 3, because:
--- a/docs/adr/adr_0012.md
+++ b/docs/adr/adr_0012.md
@@ -118,17 +118,6 @@ This fixed namespace mode has been added to the `ClusterScanCompletionHook` but 
 
 Can lead to conflicts between teams if a scan of "team 1" fails because of an error in the `ClusterScanCompletionHook` managed by a "central cluster team". Members of "team 1" will likely not even have access to the namespace the failing hooks run in and won't be able to debug / fix these issues by themselves.
 
-#### Advantages of this Approach
-
-TBD
-
-#### Disadvantages of this Approach
-
-TBD
-
-#### Security Considerations
-
-TBD
 
 ### Proposal 2: Distinct Cluster-wide and Namespace-local Resources
 
@@ -139,10 +128,12 @@ This alternative proposal uses the same cluster-wide CRDs mentioned above, but a
 The `ClusterScan` is almost identical to the `Scan` type, however, it includes an additional field called `executionNamespace` that controls in which namespace it is scheduled. The operator will schedule it in to the namespace, or throw an error if the namespace does not exist or the operator cannot schedule into it for any reason. They will only trigger a `ClusterScanCompletionHook`, only respect `ClusterCascadingRule`, and in all other ways be kept separate from non-Cluster-resources, with one major exception: access to namespace-specific ConfigMaps and Secrets. Here, this access is desireable, as it allows teams to customize the behavior of cluster-managed scans to their own situation (e.g., provide a cluster-wide ZAP scan with a namespace-specific authentication configuration for the microservice). The same consideration from the other proposal apply for ensuring that the secrets and configMaps are available.
 
 **OPEN QUESTION:** From a threat perspective (what happens if an unauthorized user can create cluster-wide scans and schedule them into other teams' namespaces?), it may be desireable to create a specific label or annotation for secrets and configmaps that has to be present for them to be accessible to cluster-wide scans, and to hide other such resources from them for security reasons, as a defense-in-depth measure against incorrectly configured clusters (we really need to threat model this feature). Is this something that is feasible with a k8s operator?
+=> This is possible, either in the operator, or using a validating webhook (the former being preferred).
 
 These CRDs will likely not be used a lot by human operators, but they are helpful for use with the autodiscovery feature, which can use them to schedule scans into namespaces without interfering with any scans, hooks or other features the owners of the namespace are using. Pods scheduled from scans like this should also receive a distinct label or annotation from regular scans in that namespace, to allow for fine-grained network policies (e.g., "Cluster scans are allowed to access this specific server that caches nuclei templates for them, but regular scans are not").
 
 **OPEN QUESTION:** Can people with rights on their own namespace, but not other namespaces, see ClusterScans that are scheduled in their namespace? Should they be able to?
+=> This should be blocked.
 
 #### `ClusterScanType`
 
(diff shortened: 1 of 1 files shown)


docs/adr/adr_0012.md
```diff
@@ -0,0 +1,129 @@
...
+Some features of normal scanTypes will not function normally with ClusterScanTypes, especially mounting values from ConfigMaps / Secrets as files / environment variables into your containers as the ConfigMaps / Secrets are not existing in every namespace and there is not sensible way for to roll them out with our helm chart to roll them out.
+
+When deploying cluster scans types you have to make sure that either:
+
+1. All referenced ConfigMaps / Secrets exist in all namespaces. This will likely be done by a script / operator generating these configmaps and persisting them into every namespace
+2. Remove all dependencies to ConfigMaps / Secrets from your scanTypes and move these either into the container image of the scanner, or by specifing a initContainer in the ScanType which copies these files over to volume shared with the scanner container. Some more references and discussions on ways to improve this in the future can be found in [ADR-0009: Architecture for pre-populating the file system of scanners"](./adr-0009)
+
+### `ClusterParseDefinition`
+
+Other then being cluster scoped `ClusterParseDefinition`s are identical to the existing `ParseDefinition` CRD.
```
SebieF, line 60:
```suggestion
Other than being cluster scoped, `ClusterParseDefinition`s are identical to the existing `ParseDefinition` CRD.
```

docs/adr/adr_0012.md
```diff
@@ -0,0 +1,129 @@
...
+
+When a scan job has completed and the results need to be parsed, the operator should first look for `ParseDefinition`s in the `Scan`s namespace matching the `scantype.spec.extractResults.type` / `clusterscantype.spec.extractResults.type` name configured in the ScanType.
+If that is not found the operator should look for `ClusterParseDefinition`s with the same name.
+No matter which was picked the `Job` for the parser always has to be started in the namespace of the `Scan`.
+
```
SebieF, line 65: I am wondering if there is even a need for namespace specific ParseDefinitions anymore, once cluster-wide parse definitions exist. Is there really a use case where you would parse the results of the same scanner differently, depending on the namespace?

docs/adr/adr_0012.md
```diff
@@ -0,0 +1,129 @@
...
+### `ClusterCascadingRule`
+
+Other then being cluster scoped `ClusterCascadingRule`s are identical to the existing `CascadingRule` CRD.
```
SebieF, line 68:
```suggestion
Other than being cluster scoped, `ClusterCascadingRule`s are identical to the existing `CascadingRule` CRD.
```

docs/adr/adr_0012.md
```diff
@@ -0,0 +1,129 @@
...
+
+When the cascading scans hook starts, the hook should fetch the list of both `CascadingRule`s and `ClusterCascadingRule`s and merge them together into one list which then get's evaluated against the scan just like it is now. All scans started from it have to be in the same namespace of the original `Scan`.
```
SebieF, line 70:
```suggestion
When the cascading scans hook starts, the hook should fetch the list of both `CascadingRule`s and `ClusterCascadingRule`s and merge them together into one list. The list then gets evaluated against the scan just like it is now. All scans started from it have to be in the same namespace of the original `Scan`.
```

docs/adr/adr_0012.md
```diff
@@ -0,0 +1,129 @@
...
+
+#### Deduplication of CascadingRules
+
+One potential issue with CascadingRule and ClusterCascadingRule is that in contrast to ScanType and ParseDefinition both namespaced and cluster wide types are used. This can be problematic when both the namespace as well as the cluster level have similar / equal CascadingRule installed as this could lead to both producing the same scan which would then be executed twice.
+
+Potential workarounds here would be to introduce an additional field to the CascadingRule spec which allows us to define a sort of "deduplication" / replacement mechanism.
+
+### `ClusterScanCompletionHook`
+
+`ClusterScanCompletionHook` behave mostly like their namespaced counterpart, when the parser completes the operator fetches the list of ScanCompletionHooks from Scans namespace (respecting the hookSelector of the Scan) and the `ClusterScanCompletionHook`s (also respecting the hookSelector of the Scan (this might have to be discussed...)) are merged and ordered according to the respective priorities set on them.
```
SebieF, line 80:
```suggestion
`ClusterScanCompletionHook` behave mostly like their namespaced counterpart, when the parser completes, the operator fetches the list of ScanCompletionHooks from Scans namespace (respecting the hookSelector of the Scan) and the `ClusterScanCompletionHook`s (also respecting the hookSelector of the Scan (this might have to be discussed...)) are merged and ordered according to the respective priorities set on them.
```

docs/adr/adr_0012.md
```diff
@@ -0,0 +1,129 @@
...
+
+The execution of the hooks is different to the current model, as there be an additional setting on the `ClusterScanCompletionHook` called namespaceMode which allows users to configure to either run the hook in the same namespace as the scan it is getting executed for, or to provide a fixed namespace where all executions for this `ClusterScanCompletionHook` will be run in.
```
SebieF, line 82:
```suggestion
The execution of the hooks is different to the current model, as there must be an additional setting on the `ClusterScanCompletionHook` called namespaceMode which allows users to configure to either run the hook in the same namespace as the scan it is getting executed for, or to provide a fixed namespace where all executions for this `ClusterScanCompletionHook` will be run in.
```

docs/adr/adr_0012.md
```diff
@@ -0,0 +1,129 @@
...
+
+```yaml
+apiVersion: "execution.securecodebox.io/v1"
+kind: ClusterScanCompletionHook
+metadata:
+  name: slack-notification
+spec:
+  type: ReadAndWrite
+  image: "docker.io/securecodebox/notification"
+```
```
SebieF, line 94: Is this yaml example necessary? It does not specify the execution namespace. If it is meant to be an example for the same namespace as the scan, I would add a comment to clarify that.

docs/adr/adr_0012.md
```diff
@@ -0,0 +1,129 @@
...
+spec:
+  # spec.executionNamespace - would let you set a fixed namespace for the hook to get executed in
+  executionNamespace: example-namespace-name
+  type: ReadAndWrite
+  image: "docker.io/securecodebox/notification"
+```
+
+This fixed namespace mode has been added to the `ClusterScanCompletionHook` but not for the `ClusterScanType` for the following reasons:
+
+1. Hooks are usually used to interface with 3rd party systems which are often standardized for the entire company. Allowing to run them in a namespace separate from the individual teams / scans allows the hook access these systems without having to distribute and thus potentially compromise the credentials for the 3rd party systems.
```
SebieF, line 110:
```suggestion
1. Hooks are usually used to interface with 3rd party systems which are often standardized for the entire company. Allowing to run them in a namespace separate from the individual teams / scans allows the hook to access these systems without having to distribute and thus potentially compromise the credentials for the 3rd party systems.
```

docs/adr/adr_0012.md
```diff
@@ -0,0 +1,129 @@
...
+2. In contrast to scans, hooks usually do not need to send requests to the scanned applications. Scanning a app from outside their namespace opens up a lot of potential issues with network policies blocking inter network traffic between scanner and app. As the hooks only need to access the S3 bucket and the 3rd party system they are interfacing with, this is not a issue for hooks.
```
SebieF, line 111:
```suggestion
2. In contrast to scans, hooks usually do not need to send requests to the scanned applications. Scanning an app from outside its namespace opens up a lot of potential issues with network policies blocking inter network traffic between scanner and app. As the hooks only need to access the S3 bucket and the 3rd party system they are interfacing with, this is not an issue for hooks.
```

docs/adr/adr_0012.md
```diff
@@ -0,0 +1,129 @@
...
+
+#### Potential Issues With the Fixed `executionNamespace` Mode
+
+Can lead to inter team conflicts if a scans of "team 1" fails because of a error in the `ClusterScanCompletionHook` managed by a central cluster team. Members of "team 1" will likely not even have access to the namespace the failing hooks run in and wont be able to debug / fix these issues by their self.
```
SebieF, line 115:
```suggestion
Can lead to conflicts between teams if a scan of "team 1" fails because of an error in the `ClusterScanCompletionHook` managed by a "central cluster team". Members of "team 1" will likely not even have access to the namespace the failing hooks run in and won't be able to debug / fix these issues by themselves.
```

SebieF (changes requested):
Very well laid out proposal :) 

Some comments might be very minor, but will hopefully help most readers to follow your arguments a bit better.

SebieF (dismissed):

malexmave:
I just added a second proposal to the ADR, which features stricter separation between cluster-wide resources and namespace-local ones. I will discuss this with @J12934 to complete the "TBD" parts before we can bring this to the rest of the team for discussion.

Issue #1250, opened 2022-06-27
Empty Screenshot or 0 Bytes PNG File


mjagyo:
Hi,

May I ask for any setting or reason why we are getting an empty screenshot that was being uploaded to Minio?

Here is the yaml file:
```
apiVersion: "execution.securecodebox.io/v1"
kind: Scan
metadata:
  name: "screenshot-github-securecodebox.test.io"
spec:
  scanType: "screenshooter"
  parameters:
    - "https://github.com/secureCodeBox"
```

Here is the file that was being uploaded to the Minio:
![image](https://user-images.githubusercontent.com/13362819/175960745-55489b1e-3528-4a7d-9937-5e4a7a2aec63.png)

Also an empty findings.json (shown above)

Thank you so much.

J12934:
Hi @mjagyo 

not aht is not the expected / wanted behavior, it seems like firefox changed their install method some time back and we now have to change the way we install it in the screenshooter container image.

Currently firefox is simply outputting:

```
Command '/usr/bin/firefox' requires the firefox snap to be installed.
Please install it with:

snap install firefox
```

We will take a look and see how we can get this fixed.

rseedorff:
Looks like the docker image needs a refactoring 🤔 Since Ubuntu 22.4 it is no longer possible to install Firefox via apt package 📦. Tried snap but that's not working within a docker container.

PR #1277, opened 2022-07-19
Fix screenshooter

the-simmon:
This PR fixes https://github.com/secureCodeBox/secureCodeBox/issues/1250

I changed the docker image to debian as it doesnt use snaps. I also added a simple integration test.

### DIFF
--- a/scanners/screenshooter/integration-tests/screenshooter.test.js
+++ b/scanners/screenshooter/integration-tests/screenshooter.test.js
@@ -14,6 +14,7 @@ test(
       60 * 4
     );
 
-    expect(categories).toBeDefined();
+    expect(categories).toEqual({"Screenshot":1});
   }, 60*1000
 );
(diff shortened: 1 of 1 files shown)


scanners/screenshooter/integration-tests/screenshooter.test.js
```diff
@@ -0,0 +1,19 @@
...
+  "make screenshot of nginx demo target",
+  async () => {
+    const { categories } = await scan(
+      "demo-target-screenshot",
+      "screenshooter",
+      ["http://nginx.demo-targets.svc"],
+      60 * 4
+    );
+
+    expect(categories).toBeDefined();
```
J12934, line 17: I think a proper expect to check if the actual screenshot category here would be better.
It could be defined but empty-

J12934 (approved):

PR #1329, opened 2022-08-30
Auto discovery consistent behavior

### LINKED ISSUE #1112: Autodiscovery consistent behavior
Currently the service and container autodiscovery have slightly different behavior. We should define what the expected behavior of both should be.

#### Manually deleting a scheduled scan
When a scheduled scan that was created by the autodiscovery gets deleted the service autodiscovery will automatically create a new scheduled scan. The container autodiscovery will not create a new scheduled scan (although when a pod is updated or a new pod with the same container gets created the autodiscovery logic will be triggered which will create a new scheduled scan). This difference in behavior is due to the fact that the service autodiscovery requeues itself (see passiveReconcileInterval config) whereas the container autodiscovery will not requeue itself at all.

#### Check if scantype is present
The service autodiscovery checks if the scantype specified in the config is installed and if not it will requeue itself until the scantype is installed. The container autodiscovery does not check if the scantype is installed. It will create a scheduled scan regardless.

#### Default annotations
The service autodiscovery defines some default annotations (defined in values.yaml), the container autodiscovery does not define any annotations in its default config.

the-simmon:
This PR fixes #1112 
Additionally I enabled the container autodiscovery by default.


J12934 (changes requested):
please add upgrading notes to the UPGRADING.md :)
Otherwise i think this should be good to go 👍

Ilyesbdlala (approved):

Ilyesbdlala (approved):

PR #1335, opened 2022-09-01
ADR-0013: Autodiscovery v2

malexmave:
Adding a draft version of ADR-0013, which specifies options for changes to the Autodiscovery functionality to make it more flexible. Review, extensions and feedback appreciated.


the-simmon (dismissed):

Weltraumschaf:
My 50 cent: From architecture perspective option 1 has a major drawbacl/consequnce missed in the ADR. It introduces a new concept of flow control via events. Flow control via events are very flexible, but comes with the drawback that the flow control is hard to understand and trace because it is not obvious from the static code. This makes debugging and reasoning of control flow harder. In consequence this adds cost for later maintenance of the system. So this should be taken carefully into account.

malexmave:
I've added your concern. We can merge this now and create a new PR when we want to actually document a decision.

Weltraumschaf (approved):

PR #1336, opened 2022-09-01
ADR-0014: Metrics Collection

malexmave:
First draft for the metrics collection ADR. Discussions welcome.


J12934:
awesome 👍

A couple of notes while going over the adr:

1. Parsers don't save the finding into a file in the filesystems but return it directly as a function return value to the parser-sdk which then converts these into json and uploads them to s3. The points you made are still valid in that scenario :)
2. A lot of scanners, including ZAP (i think) don't put this kind of information into their result file. So the parser won't always have the information available to properly gather these metrics. I think we could come up with additions to this pr to allow scanners to report additional metrics via the lurker, e.g. by using an additional (optional) file as suggested for the parser. This could then be directly generated by a scanner (e.g. zap-advanced) or produced by a special SCB specific sidecar collecting these metrics and writing them.

Weltraumschaf:
LGTM

malexmave:
Incorporated the feedback from @J12934. We can now review and merge this as a WIP ADR so we can share it with others to get their feedback on it.

Weltraumschaf (approved):

Weltraumschaf:
Since it's a DRAFT we can merge it.

PR #1340, opened 2022-09-03
Upgraded embedded minio chart to latest release v11.9.4 (closes #755)

### LINKED ISSUE #755: Update Minio Version used in Operator Helm Chart

The minio version included in the operator by default is quite old. Minio themself now only rarely update [their helm chart](https://artifacthub.io/packages/helm/minio/minio), they are now recommending to use their operator. The minio operator is not really a option for us as we can't rely on that the operator is installed.

There is an alternative helm chart provided by bitnami which looks like it still gets regular update: https://artifacthub.io/packages/helm/bitnami/minio

We could switch over to the bitnami chart, but this would likely require a major release as they use different helm values for configuration.

rseedorff:
## Description

Upgraded embedded minio chart to latest release v11.9.4 (closes #755)

### Checklist


operator/templates/manager/manager.yaml
```diff
@@ -75,14 +75,14 @@ spec:
               valueFrom:
                 secretKeyRef:
                   name: "{{ .Release.Name }}-minio"
-                  key: accesskey
+                  key: root-user
```
J12934, line 78: This will also require tweaks to the installation docs: https://www.securecodebox.io/docs/getting-started/installation#accessing-the-included-minio-instance

operator/Chart.lock
```diff
@@ -1,6 +1,6 @@
 dependencies:
 - name: minio
-  repository: https://helm.min.io/
-  version: 7.1.2
-digest: sha256:6fee974baf705caae258636dbd7b2b89f14230b9ae28aa20da4643d8292c399b
-generated: "2020-10-07T11:26:59.229968+02:00"
+  repository: https://charts.bitnami.com/bitnami
+  version: 11.9.4
+digest: sha256:dcd5c66891556818e8955c7aabd358ca8c87928291b1918611cb333184dd5109
+generated: "2022-09-03T17:30:31.769197+02:00"
```
the-simmon, line 6: Is there are reason why you use the bitnami chart instead of the official chart mentioned [here](https://github.com/secureCodeBox/secureCodeBox/issues/755#issuecomment-1123256675)? The discussion never came to a final conclusion on what to use.

PR #1342, opened 2022-09-05
Add Resources to Scan, ParseDefinition & ScanCompletionHook CRDs

### LINKED ISSUE #1327: Allow setting resource limits for pods through Hook and Parser CRDs / values.yaml

The hook and parser CRDs currently do not allow setting resource limits (unlike the ScanType CRD, where it is supported). Since people are running into issues with that (#1305, #1320), we should fix this :)

Relevant part of the operator code [for parser](https://github.com/secureCodeBox/secureCodeBox/blob/main/operator/controllers/execution/scans/parse_reconciler.go#L143-L150) and [for hook](https://github.com/secureCodeBox/secureCodeBox/blob/main/operator/controllers/execution/scans/hook_reconciler.go#L365-L371).

This will require changes to the CRDs, but they will not be breaking, so this could be done independently from a v4 release.

J12934:
Closes #1327 

## Open Todo's

- [x] Add resource values to ALL scanner charts (f**or parsers**)
- [x] Add resource values to ALL hook charts
- [x] Update API Docs

Scanner Example Default HelmValues:
```yaml
parser:
  # parser.resources -- CPU/memory resource requests/limits (see: https://kubernetes.io/docs/tasks/configure-pod-container/assign-memory-resource/, https://kubernetes.io/docs/tasks/configure-pod-container/assign-cpu-resource/)
  resources: {}
  #   resources:
  #     requests:
  #       memory: "256Mi"
  #       cpu: "250m"
  #     limits:
  #       memory: "512Mi"
  #       cpu: "500m"
```
Parser Example Default HelmValues:
```yaml
hook:
  # hook.resources -- CPU/memory resource requests/limits (see: https://kubernetes.io/docs/tasks/configure-pod-container/assign-memory-resource/, https://kubernetes.io/docs/tasks/configure-pod-container/assign-cpu-resource/)
  resources: {}
  #   resources:
  #     requests:
  #       memory: "256Mi"
  #       cpu: "250m"
  #     limits:
  #       memory: "512Mi"
  #       cpu: "500m"
```


operator/controllers/execution/scans/parse_reconciler.go
```diff
@@ -91,6 +91,21 @@ func (r *ScanReconciler) startParser(scan *executionv1.Scan) error {
...
+	resources := corev1.ResourceRequirements{
+		Requests: corev1.ResourceList{
+			corev1.ResourceCPU:    resource.MustParse("200m"),
+			corev1.ResourceMemory: resource.MustParse("100Mi"),
+		},
+		Limits: corev1.ResourceList{
+			corev1.ResourceCPU:    resource.MustParse("400m"),
+			corev1.ResourceMemory: resource.MustParse("200Mi"),
+		},
+	}
```
the-simmon, line 104: Wouldnt it be easier to use default values in the CRD definition instead of hardcoding it in code?
https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting
This is also true for all the other hardcoded default resource limits in other files.

rseedorff: There is a way to define defaults with kubebuilder: https://book.kubebuilder.io/reference/markers/crd-validation.html

J12934: yes good point that would indeed be better.

I've had problems with the kubebuilder defaults before they haven't worked consistently for me before but we can definitely give it another try. Might be a bit hard in this case thought as we don't define the individual fields ourself but pull them in from the kubernetes api types.

J12934: Applied the defaults working nicely 👍
Unfortunately the defaults are only applied when the custom resources are applied, so just doing it with the defaults would break existing scans. So we will to use both methods until v4 to make sure existing installations with upgraded operators keep working.

PR #1345, opened 2022-09-07
Remove redundant job type annotations

### LINKED ISSUE #1244: Remove redundant job-type annotations
Currently we have a `job-type` **label** and **annotation**. 

The `job-type` **annotation** for hooks, parsers and scans should be removed in the next breaking release because the annotations are redundant.

https://github.com/secureCodeBox/secureCodeBox/blob/0f68d158b14e69247cef33b2339debba71776d9b/operator/controllers/execution/scans/parse_reconciler.go#L109-L113

https://github.com/secureCodeBox/secureCodeBox/blob/0f68d158b14e69247cef33b2339debba71776d9b/operator/controllers/execution/scans/hook_reconciler.go#L347-L351

https://github.com/secureCodeBox/secureCodeBox/blob/0f68d158b14e69247cef33b2339debba71776d9b/operator/controllers/execution/scans/scan_reconciler.go#L200

the-simmon:
This PR fixes #1244

### DIFF
--- a/operator/controllers/execution/scans/hook_reconciler.go
+++ b/operator/controllers/execution/scans/hook_reconciler.go
@@ -347,7 +347,6 @@ func (r *ScanReconciler) createJobForHook(hook *executionv1.ScanCompletionHook, 
 					Annotations: map[string]string{
 						"auto-discovery.securecodebox.io/ignore": "true",
 						"sidecar.istio.io/inject":                "false",
-						"securecodebox.io/job-type":              "hook",
 					},
 				},
 				Spec: corev1.PodSpec{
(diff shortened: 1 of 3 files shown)


malexmave:
@J12934 Just double-checking: is removing the annotations fine? I dimly recall that certain types of filtering in Kubernetes can only be done on annotations, not labels (e.g. for network policies etc.), but according to [this source](https://blog.getambassador.io/kubernetes-labels-vs-annotations-95fc47196b6d), labels are the thing that is used there, and annotations aren't used by the system, so it seems to be the other way around.

J12934:
> @J12934 Just double-checking: is removing the annotations fine? I dimly recall that certain types of filtering in Kubernetes can only be done on annotations, not labels (e.g. for network policies etc.), but according to [this source](https://blog.getambassador.io/kubernetes-labels-vs-annotations-95fc47196b6d), labels are the thing that is used there, and annotations aren't used by the system, so it seems to be the other way around.

yes it's fine to remove.

I don't remember the discussion but annotations are never used for selection (to my knowledge). Labels are indexed and have a severe limit on the text lengt and allowed characters. Annotations can be way longer and support most characters.

J12934 (approved):

PR #1351, opened 2022-09-10
Upgrade Operator & AutoDiscovery to latest kubebuilder 3.6.0 dependencies

J12934:
## Description

This PR updates the operator and auto-discovery to use the same dependencies as new projects scaffolded with kubebuilder 3.6.0 and replaces the scaffolded make targets with the new verions.

This allows us to upgrade the used go version to 1.19.

The PR should not be breaking would still put this in the 4.0.0 release rather then the 3.15.0 just to be sure.
Note: kubebuilder 3.6.0 / its go/v3 plugin is still not compatible with apple silicon machines as its using a older version of kustomize. The go/v4 plugin which fixes this is not stable so we're still using the go/v3 plugin. See: https://book.kubebuilder.io/quick-start.html#create-a-project

### Checklist


Ilyesbdlala (approved):
LGTM. AFAIK no upgrade notes needed

PR #1389, opened 2022-10-05
Generate S3 URLs from go template

### LINKED ISSUE #1137: How to add a suffix to output files
## Your Question

I got two scheduled scans to two different sites which appear in the cluster as:
```
NAME                        TYPE            INTERVAL   FINDINGS
zap-scan-earnd-admin        zap-full-scan   168h       19
zap-scan-egw-adminconsole   zap-full-scan   168h       21
```

The scans generate pods like these ones.
```
scan-zap-scan-earnd-admin-1650962558-p77qz-dhn6x
scan-zap-scan-egw-adminconsole-1650962558-82c24-m6zks 
```

However, when the scan results are written to the S3 bucket, we get folders like:
```
scan-202639fa-0083-42b0-a401-91b167c44370/
scan-da96b881-9b2f-4b9c-93da-afe3d315ce5c/
```

There is no way to identify which one is the result of a particular scan or even from a particular run.

What I would like to see is the folders in S3 like this:
```
zap-scan-earnd-admin-2022-04-03-12-35/
zap-scan-earnd-admin-2022-04-10-12-35/
zap-scan-egw-adminconsole-2022-04-03-12-35/
zap-scan-egw-adminconsole-2022-04-10-12-35/
```

That is, I would like to see the folder named relative to the scan name and the date/time when it was run.

Is there a way to do this?

Thanks for your help.

the-simmon:
fixes #1137

This PR adds a config option to the operator where the user can provide a go template to change the default filenames for s3. The default itself did not change compared to previous operators.

### DIFF
--- a/operator/controllers/execution/scans/hook_reconciler.go
+++ b/operator/controllers/execution/scans/hook_reconciler.go
@@ -198,12 +198,12 @@ func (r *ScanReconciler) processPendingHook(scan *executionv1.Scan, status *exec
 	}
 
 	var rawFileURL string
-	rawFileURL, err = r.PresignedGetURL(scan.UID, scan.Status.RawResultFile, defaultPresignDuration)
+	rawFileURL, err = r.PresignedGetURL(*scan, scan.Status.RawResultFile, defaultPresignDuration)
 	if err != nil {
 		return err
 	}
 	var findingsFileURL string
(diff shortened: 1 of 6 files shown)


J12934 (approved):
looks great 👍

PR #1427, opened 2022-11-01
Docs: Make separation between cluster- and namespace modes more explicit in ADR12 (Cluster-wide resources)

malexmave:
We found that the separation in the description of one of the proposals for the cluster-wide resources was insufficiently clear. With this PR, I'm adding some more explicit notes in some places.


Ilyesbdlala (approved):
LGTM

PR #1446, opened 2022-11-17
Added new DNS Scanner: Doggo

rseedorff:
## Description

Added a new DNS Scanner: Doggo 

### Checklist


scanners/doggo/examples/example.com/scan.yaml
```diff
@@ -0,0 +1,14 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+apiVersion: "execution.securecodebox.io/v1"
+kind: Scan
+metadata:
+  name: "doggo-localhost"
+spec:
+  scanType: "doggo"
+  parameters:
+    - "-invalidFlag"
```
J12934, line 12: examples should be working and not include invalid flags :)

scanners/doggo/parser/parser.js
```diff
@@ -0,0 +1,70 @@
...
+
+  return result;
+}
+
+function transformToFindings(targets) {
+  // Code to transform the scanner results to scb findings
+  return targets.map((target) => {
+    return {
+      name: `DNS Zone: ${target.name} | Type: ${target.type}`,
+      description: `DNS record type "${target.type}" found for zone "${target.name}" with address "${target.address}".`,
```
J12934, line 28: Is "zone" here really the correct / most helpful term?
Aren't these DNS records in a zone?

J12934 (commented):
Look good 👍

I've added some missing some example findings and refactored the parser a bit to be more in line with our other parsers.
I got one question in the comments below which would be nice to clear up before merging

scanners/doggo/parser/parser.js
rseedorff: Thx for the hint. Now fixed as discussed 👍🏻

Weltraumschaf (approved):

PR #1456, opened 2022-11-22
Constrints vs solution strategy

Weltraumschaf:
The ADRs about k8s and s3


J12934 (changes requested):
wrote down the reasons for s3

J12934 (approved):

PR #1526, opened 2023-01-02
Add Apikey in generic webhook

srburton:
### Checklist


hooks/generic-webhook/hook/hook.js
```diff
@@ -8,13 +8,17 @@ async function handle({
...
+  webhookApikeySecret = process.env["WEBHOOK_APIKEY_SECRET"],
   axios = require('axios')
 }) {
   const findings = await getFindings();
 
   console.log(`Sending ${findings.length} findings to ${webhookUrl}`);
 
-  if (webhookUser && webhookPassword){
+  if (webhookApikey && webhookApikeySecret){
+    await axios.post(webhookUrl, {scan, findings }, {headers: { webhookApikey: webhookApikeySecret}});
```
J12934, line 20: I don't think this works as intended as the key would always be hardcoded to `webhookApikey` not the value of the var.

I think it should work like this.

```suggestion
    await axios.post(webhookUrl, {scan, findings }, {headers: { [webhookApikey]: webhookApikeySecret}});
```

hooks/generic-webhook/templates/webhook-hook.yaml
```diff
@@ -31,6 +31,18 @@ spec:
           name: {{ .Values.hook.authentication.basic.userSecret }}
           key: {{ .Values.hook.authentication.basic.passwordKey }}
           optional: true
+    - name: WEBHOOK_APIKEY
+      valueFrom:
+        secretKeyRef:
+          name: {{ .Values.hook.authentication.apikey.userSecret }}
+          key: {{ .Values.hook.authentication.apikey.key }}
```
J12934, line 38:
```suggestion
          key: header-name
```

hooks/generic-webhook/templates/webhook-hook.yaml
```diff
@@ -31,6 +31,18 @@ spec:
...
+          optional: true
+    - name: WEBHOOK_APIKEY_SECRET
+      valueFrom:
+        secretKeyRef:
+          name: {{ .Values.hook.authentication.apikey.userSecret }}
+          key: {{ .Values.hook.authentication.apikey.secret }}
```
J12934, line 44:
```suggestion
          key: header-value
```

hooks/generic-webhook/values.yaml
```diff
@@ -35,8 +35,14 @@ hook:
...
 
-  # hook.authentication -- Optional basic authentication credentials
+  # hook.authentication -- Optional basic authentication credentials or apikey
   authentication:
+    apikey:
+      # -- Link a pre-existing generic secret with `usernameKey` and `passwordKey` key / value pairs
+      userSecret: generic-webhook-credentials
+      # -- Customize header name as per your needs
+      key: X-api-key
+      secret: 26ea529e517748baa6d87ebfe5781475
```
J12934, line 45: This also doesn't work as intended.
The key / secret should not be specified here using the helm values but should be put in a pre-existing secret.
So the user creates the secret beforehand and then only links to the secret from the helm values. I've noted how remove them in the templates above.

J12934 (changes requested):
Hi, thank you for the PR 🙌🚀

I think there a a couple of errors which would prevent this from working, I've noted them down in the review below.

Can you please also take a look at the DCO requirements listed here: https://github.com/secureCodeBox/secureCodeBox/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-and-licensing

I can't merge PRs without it.

hooks/generic-webhook/hook/hook.js
renatosazup: @J12934  
In axios headers are implemented this way.

```javascript

const result = await axios.post('https://testapi.org/post', { name: 'John Doe' }, {
    headers: {
    'content-type': 'text/json'
    }
});

```
### Reference
https://stackabuse.com/how-to-send-headers-with-an-axios-post-request/

J12934: @renatosazup yes, the problem is that you don't set the key properly but always send along the string "webhookApikey" along as the header name because you are not using computed property names but just a string there: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#computed_property_names

hooks/generic-webhook/values.yaml
srburton: Sorry @J12934 I don't understand.

srburton: @J12934, check my corrections please?

rseedorff:
Hi @renatosazup,
thanks for your question. We plan to release v4.0.0 of secureCodeBox in the next weeks 🤞
Can we merge your PR already, just asking because its marked as draft, but seems to be ready.

renatosazup:
@rseedorff you can merge yes, it's ready 😄

rseedorff (approved):

PR #1557, opened 2023-01-21
Enable ContainerAutodiscovery to scan images from private repos (ADR 17)(closes #1189)

### LINKED ISSUE #1189: Enable Container AutoDiscovery to work with non public images

The container Auto Discovery currently only works with public images only. Scans created by it for non public images will fail.
This should be changed so that it uses the pull secrets specified in the pod & the pull secrets specified in the ServiceAccount to be able to access these image.

Ideally this should be implemented in a way which doesn't involve the auto discovery actually fetching the secrets as this would be a very high privileged right.

- [x] Document ADR @the-simmon, #1374

the-simmon:
fixes #1189


malexmave:
The Python part looks good to me.

J12934 (approved):

Issue #519, opened 2021-06-24
Improve Findings Format


JohannesZahn:
The [Findings Format](https://docs.securecodebox.io/docs/api/finding) is used e.g. to export data to other Applications like DefectDojo, however some necessary and/or useful information is not included in the Findings. 

Therefore we will collect a List of possible Improvements here. We are happy about any suggestions by the community.

# Release 4.0 Scope
- [x] Add Timestamp of when a Finding was identified (called `identified_at`), it should be optional as we do not always have this exact information. #1434

- [X] Add Timestamp of when a Finding was parsed (called `parsed_at`) as a fallback when `identified_at` is not present. Will be solved by #492 
- [ ] Add additional JSON attributes, that give information about how a Finding was identified ("evidence", "steps_to_reproduce"?) or what impact it has ("impact"?), see **https://github.com/secureCodeBox/secureCodeBox/issues/332#issuecomment-835778057** @secureCodeBox/contributer-team do we have any data available to do this?
- [x] Add optional JSON attribute "mitigation" for solutions how the vulnerability might be fixed, can be populated i.e. by "zap_solution" in zap  https://github.com/secureCodeBox/secureCodeBox/pull/1639
- [x]  Make severity, category and name required **breaking** => update findings format check 
- [ ] Add optional JSON attribute "cve" for CVEas present i.e. in trivy https://github.com/secureCodeBox/secureCodeBox/pull/1676
- [ ] Add optional JSON attribute "cwe" for CWE as present i.e. in ZAP https://github.com/secureCodeBox/secureCodeBox/pull/1676
- [ ] Add optional JSON attribute "notes"/"other"/"additional_info" for other important information text that clarifies the vulnerability. 

# Additional Ideas for further development
- [ ] Add optional JSON attribute "vulnerability_id_from_tool" for ids of vulnerabilities assigned by the specific scanners

@secureCodeBox/contributer-team please share any ideas you have in the comments.

JohannesZahn:
@damiencarol when you were talking about the Request/Answer Part [here](https://github.com/secureCodeBox/secureCodeBox/issues/332#issuecomment-835778057), do those resemble the mitigation and impact fields in your Findings format or what other fields would be populated with this information?

JohannesZahn:
Some Scanners like nikto seem to supply vulnerability IDs like in the following examle. Could this be useful information for the "vulnerability_id_from_tool" field in DefectDojo @damiencarol?

```json
[
  {
[...]

PR #1639, opened 2023-03-13
Added optional `mitigation` attribute to findings

Ilyesbdlala:
## Description

Relates to #519 
The attribute `mitigation` is added to scanners who include explicit solution in their results or where the solution is almost always the same (i.e ncrack -> use a more secure password)
Scanners that are usually informational (such as NMAP) are excluded.

Scanners affected:

- ncrack
- ssh-scan 
- trivy
- typo3scan
- Zap/Zap-Advanced
- sslyze

### Checklist


UPGRADING.md
```diff
@@ -179,3 +179,7 @@ One of the optional attributes in the finding format of Amass has been renamed f
 ➡️  [Reference: #1605](https://github.com/secureCodeBox/secureCodeBox/pull/1605)
 
 
+### Renamed the scanners' ssh-scan and sslyze  `hint` to `mitigation`
```
J12934, line 182:
```suggestion
### Renamed the scanners ssh-scan and sslyze  `hint` to `mitigation`
```

scanners/ssh-scan/parser/parser.js
```diff
@@ -180,7 +180,7 @@ async function parse(fileContent) {
         osi_layer: "APPLICATION",
         severity: "INFORMATIONAL",
         reference: {},
-        hint: "",
+        mitigation: "",
```
J12934, line 183: It would be nicer if these could always be null when not set instead of an empty string

J12934 (changes requested):
Looking good, some really minor notes then it should be ready to go 🚀

J12934 (dismissed):

J12934 (approved):

PR #1645, opened 2023-03-15
Enforced the use of the node version in .nvmrc

### LINKED ISSUE #1581: Respect .nvmrc in Workflows
Long time ago we introduced `.nvmrc` to manage the used nodejs/npm versions. But we use special vars for that in our workflow files:

```
.github/workflows/ci.yaml
24:  NPM_VERSION: "7"
199:        run: npm i -g npm@${{ env.NPM_VERSION }}
325:        run: npm i -g npm@${{ env.NPM_VERSION }}

.github/workflows/release-build.yaml
16:  NPM_VERSION: "7"
```

We should consider using nvm instead.

Ilyesbdlala:
## Description

closes #1581 

### Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -163,6 +163,20 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: '${{ env.PYTHON_VERSION }}'
+
+      - name: Download nvm and Node.js
+        uses: actions/download-artifact@v2
+        with:
+          name: nvm-node
+          path: nvm-node
+
+      - name: Extract nvm and Node.js
+        run: |
+          tar xzf nvm-node/nvm-node.tar.gz -C $HOME
+          export NVM_DIR="$HOME/.nvm"
+          [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
+          nvm use      
+          
       - name: Unit Tests
         working-directory: ./auto-discovery/kubernetes/pull-secret-extractor
         run: make unit-test
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -155,6 +155,7 @@ jobs:
   auto-discovery-kubernetes-secret-extraction-container:
     name: "Autodiscovery | Kubernetes | SecretExtractionInitContainer"
     runs-on: ubuntu-20.04
+    needs: nvm-setup
     steps:
       - name: Checkout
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -20,10 +20,35 @@ env:
...
   GO_VERSION: "1.19"
   PYTHON_VERSION: "3.8.10"
-  NODE_VERSION: "16"
-  NPM_VERSION: "8.19.2"
+  NVM_VERSION: "0.39.3" # Does not need to be updated as often as the Node.js version or the NPM version
 
 jobs:
+
+  # ---- Configuration ----
+  config:
```
Weltraumschaf, line 28: We should name thos rather nvm-setup or similar because "config" is way to generic.

.github/workflows/ci.yaml
```diff
@@ -20,10 +20,35 @@ env:
...
+      - uses: actions/checkout@v2
+
+      - name: Install nvm and Node.js
+        run: |
+          curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v${{ env.NVM_VERSION }}/install.sh | bash
+          export NVM_DIR="$HOME/.nvm"
+          [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
+          nvm install "$(cat .nvmrc)"
+          
+      - name: Archive nvm and Node.js
```
Weltraumschaf, line 41: What is thepurpose of archiving and uploading? Pleas leave a comment in the action why this is done.

Weltraumschaf (changes requested):

.github/workflows/ci.yaml
Ilyesbdlala: My reasoning here was that this job can later be used for any "pre-staging" that we would need. So I chose a generic name. But maybe this is a case of YAGNI.

.github/workflows/ci.yaml
Ilyesbdlala: It seemed more efficient to do the installation process of nvm only once and upload it as an archive. This can later be imported and extracted quickly in the hooks/scanners testing containers. It also ensures that we pull from  an external source to install nvm only once, instead of _**x**_ number of hooks + _**y**_ number of scanners

Weltraumschaf: You should write exactly this as a comment into the code.

Weltraumschaf (approved):

PR #1679, opened 2023-04-09
Improve Nmap Parser to handle multiple / ipv6 addresses and verbose output

### LINKED ISSUE #1678: Nmap Findings include only one **ipv4** address
### Describe the bug

The current state of the nmap parser only considers ipv4 address and will only include one even if nmap has identified more then one address for the host

### Steps To Reproduce

Start the following scan (cluster must be ready to handle ipv6 traffic)

```yaml
kind: Scan
metadata:
  name: "nmap-scanme.nmap.org"
spec:
  scanType: "nmap"
  parameters:
    - "-6" # force ipv6 mode
    - scanme.nmap.org
```

Findings won't have a ip address set.

### Expected behavior

Include all ips for a host in the finding (including v4 & v6)

 - secureCodeBox Version/Release: any

J12934:
## Description

Closes #1678

### Checklist


UPGRADING.md
```diff
@@ -208,3 +208,25 @@ Example config:
...
+  "severity": "INFORMATIONAL",
+  "osi_layer": "NETWORK",
+  "attributes": {
+-    "ip_address": "10.102.186.25",
++    "ip_addresses": ["10.102.186.25"],
+    "hostname": "juice-shop.demo-targets.svc.cluster.local",
+    "operating_system": null
+  }
+}
+```
```
Ilyesbdlala, line 232: Can you add a test case where the results include multiple IP addresses at the same time i.e something like this:
```xml
<address addr="2600:3c01::f03c:91ff:fe18:bb2f" addrtype="ipv6"/>
<address addr="192.1.2.3" addrtype="ipv4"/>
```
As far as I can tell, this is not possible, nmap would either output the ipv4 or ipv6 address. So it does not make sense to have `ip_addresses` as an array. Unless I am missing something.

Ilyesbdlala (approved):

PR #1701, opened 2023-04-21
Fix hostOrIP helper in CascadingScans

J12934:
Looking at this mess, i guess the better thing would be to clean this up (by standardizing ips like we did with hostname) before v4?


hooks/cascading-scans/hook/hook.ts
```diff
@@ -216,29 +277,74 @@ function mergeCascadingRuleWithScan(
...
+    return finding.attributes["ip"];
+  }
+  if (
+    finding.attributes["ip_addresses"] &&
+    isArray(
+      finding.attributes["ip_addresses"] &&
+        finding.attributes["ip_addresses"].length > 0
+    )
+  ) {
+    return finding.attributes["ip_addresses"][0];
```
Ilyesbdlala, line 342: A comment referencing https://github.com/secureCodeBox/secureCodeBox/pull/1679 here would be nice. Because it seems like we're prioritizing the first IP address.

Ilyesbdlala (approved):
LGTM, apart from a small suggestion.

hooks/cascading-scans/hook/hook.ts
J12934: That is because we do :D
The helper needs to pick out one address. I changed it so that it always consistently picks out the first address sorted alphabetically so that the address is consistent and doesn't depend on the order of ips in the field

rseedorff (approved):

Ilyesbdlala:
LGTM. But some scanners use `NULL` as a value for empty ip_adresses and some use `undefined`. AFAIK, it's better to use `NULL`. At least that's what I used here https://github.com/secureCodeBox/secureCodeBox/pull/1676

Weltraumschaf:
> LGTM. But some scanners use `NULL` as a value for empty ip_adresses and some use `undefined`. AFAIK, it's better to use `NULL`. At least that's what I used here #1676

I suggest to open a new PR for that, in hope we have more small short lived PRs 😬

PR #1703, opened 2023-04-25
Set example accesskey and secretkey names to same as the secret example

mikkeschiren:
Just to simplify setup for "copy paste setups"

### DIFF
--- a/operator/values.yaml
+++ b/operator/values.yaml
@@ -97,8 +97,8 @@ s3:
   keySecret: my-secret
   # Names to the attributes in the s3 secret
   secretAttributeNames:
-    accesskey: accesskey
-    secretkey: secretkey
+    accesskey: accessKey
+    secretkey: secretKey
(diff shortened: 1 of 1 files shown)


Ilyesbdlala:
Thank you for Contribution :) Can you please clarify what you mean by "copy paste setups" ?

mikkeschiren:
@Ilyesbdlala - just refering to if you use the example text in the values file - a.k.a. "copy paste" :)

mikkeschiren:
The values file has this:
```
  # Example creation via kubectl:
  # kubectl create secret generic my-secret --from-literal=accessKey="******" --from-literal="secretKey=******"
  ```
  So just changing to the same case.

Ilyesbdlala:
@mikkeschiren Has your issue been resolved? Or does the solution I suggested work for you ?

Ilyesbdlala:
We decided on fixing the comment instead of the actual attributes https://github.com/secureCodeBox/secureCodeBox/pull/1839. Hope this resolves the issue.

PR #1713, opened 2023-05-04
Added new ssh-audit scanner to replace deprecated ssh-scan scanner

Reet00:
### Checklist


scanners/ssh-audit/parser/parser.js
```diff
@@ -0,0 +1,262 @@
...
+            description: "Weak SSH message authentication code algorithms in use",
+            hint: "Change these MAC algorithms"
+        },
+        enc: {
+            name: "SSH Encryption Algorithms must be changed",
+            description: "Weak SSH encryption algorithms in use",
+            hint: "Change these encryption algorithms"
+        },
+
+        addCritical: {
```
J12934, line 94: This looks weird, the `addCritical` seems to be inside the `chgWarning` object. 
Is that really intended or a typo?

scanners/ssh-audit/parser/parser.js
```diff
@@ -0,0 +1,262 @@
...
+    var severity = "low";
+    if (recommendationSeverityLevel == "critical") severity = 'HIGH'
+    if (recommendationSeverityLevel == "warning") severity = 'MEDIUM'
+    const findingTemplate = null;
+    // recommendationAction = del/chg/add
+    Object.entries(value).map(([recommendationAction, algorithms]) => {
+        //algorithmType = kex/ key/ mac, , algorithmNames = {name+note}
+        Object.entries(algorithms).map(([algorithmType, algorithmData]) => {
+            const algorithmNames = []
+            Object.entries(algorithmData).flatMap(([keyNames, content]) => { algorithmNames.push(Object.values(content)) })
```
J12934, line 160:
```suggestion
            const algorithmNames = entries(algorithmData).flatMap(([keyNames, content]) => Object.values(content))
```

scanners/ssh-audit/parser/parser.js
```diff
@@ -0,0 +1,262 @@
...
+            var action = "";
+            if (recommendationAction == "del" && recommendationSeverityLevel == "critical") action = "delCritical"
+            else if (recommendationAction == "del" && recommendationSeverityLevel == "warning") action = "delWarning"
+            else if (recommendationAction == "chg" && recommendationSeverityLevel == "critical") action = "chgCritical"
+            else if (recommendationAction == "chg" && recommendationSeverityLevel == "warning") action = "chgWarning"
+            else if (recommendationAction == "add" && recommendationSeverityLevel == "critical") action = "addCritical"
+            else if (recommendationAction == "add" && recommendationSeverityLevel == "warning") action = "addWarning"
+            const findingTemplate = templates[action][algorithmType] || null;
```
J12934, line 168: Is there a reason to have this "action" to be build by combining `recommendationAction`and `recommendationSeverityLevel` rather than having both be a new structure in the templates object?

For example if the templates object would look like this

```js
del: {
    critical: {
        kex: {
            name: "Insecure SSH KEX Algorithms",
            ...
        },
        key: ...,
        mac: ...,
        enc: ...,
    },
    warning: {
        kex: {
            name: "Insecure SSH KEX Algorithms",
            ...
        },
        key: ...,
        mac: ...,
        enc: ...,
    },
},
```

you should be able to get the finding template from templates by referencing it by 
`templates[recommendationAction][recommendationSeverityLevel][algorithmType]` without building up this action key.

scanners/ssh-audit/parser/parser.js
```diff
@@ -0,0 +1,262 @@
...
+                findingTemplate['severity'] = severity
+                findingTemplate['category'] = "SSH Policy Violation"
+                
+                combinedAlgorithmNames = []
+                algorithmNames.map(([algName, note]) => {
+                    if (note == "") combinedAlgorithmNames.push(algName)
+                    else combinedAlgorithmNames.push((algName + " (Note: " + note + ")"))
+                })
+
+                findingTemplate['algorithms'] = combinedAlgorithmNames.flat()
```
J12934, line 180:
```suggestion
                findingTemplate['algorithms'] = algorithmNames.flatMap(
                    ([algName, note]) => note == "" ? algName : `${algName} (Note: ${note})`
                )
```

scanners/ssh-audit/parser/parser.js
```diff
@@ -0,0 +1,262 @@
...
+        findingTemplate['severity'] = severity
+        findingTemplate['cvssv2'] = cvssv2
+
+        findingTemplate['references'] = []
+        findingTemplate['references'].push({'type':'CVE', 'value':`${name}`})
+        findingTemplate['references'].push({'type':'URL', 'value':`https://nvd.nist.gov/vuln/detail/${name}`})
+        cvesFindings.push(findingTemplate)
+    })
+    return cvesFindings;
+}
```
J12934, line 215: Can also be simplified by directly using the map result and simplying the finding template by just returning it directly instead of gradually building it up.

```suggestion
function transformCVEtoFinding(cves) {
    const cvesArray = Object.values(cves)
    return Object.values(cvesArray).flatMap(({cvssv2, description, name}) => {
        let severity = 'HIGH'
        if (cvssv2 < 4) severity = "LOW"
        else if (cvssv2 < 7) severity = "MEDIUM"

        return {
            name,
            description,
            category: "SSH Violation",
            severity,
            cvssv2,
            references: [
                {type: 'CVE', value: `${name}`},
                {type: 'URL', value: `https://nvd.nist.gov/vuln/detail/${name}`}
            ],
        };
    });
}
```

J12934: At the moment this calls Object.values twice. I wasn't sure if that was intentional or a misstake

scanners/ssh-audit/parser/parser.js
```diff
@@ -0,0 +1,262 @@
+// SPDX-FileCopyrightText: the secureCodeBox authors
+//
+// SPDX-License-Identifier: Apache-2.0
+
+const templates = {
```
J12934, line 5: Really like this declarative object approach to structuring the findings 👍

scanners/ssh-audit/parser/parser.test.js
```diff
@@ -0,0 +1,587 @@
...
+          "ecdh-sha2-nistp256",
+          "ecdh-sha2-nistp384",
+          "ecdh-sha2-nistp521",
+        ],
+        "category": "SSH Policy Violation",
+        "description": "Discouraged SSH key exchange algorithms in use",
+        "hint": "Remove these KEX algorithms",
+        "name": "Insecure SSH KEX Algorithms",
+        "severity": "HIGH",
+      },
```
J12934, line 139: At the moment these findings are not really usable (at least for users of the elasticsearch provider) as they don't contain a location. 
So the users could not tell which ssh server is actually using these discouraged kex algs.
All findings should have a location set (as done for the "SSH Service" finding)

scanners/ssh-audit/parser/parser.test.js
```diff
@@ -0,0 +1,587 @@
...
+            "hmac-sha2-512",
+            "hmac-sha1",
+          ],
+          "server_banner": "SSH-2.0-OpenSSH_7.2p2 Ubuntu-4ubuntu2.8",
+          "ssh_lib_cpe": "OpenSSH_7.2p2",
+          "ssh_version": 2,
+        },
+        "category": "SSH Service",
+        "description": "Information about Used SSH Algorithms",
+        "location": "dummy-ssh.demo-targets.svc",
```
J12934, line 122: Location should be a valid URI (at least if possible) so this should start with the protocol.
So: `ssh://dummy-ssh.demo-targets.svc`

scanners/ssh-audit/parser/parser.test.js
```diff
@@ -0,0 +1,587 @@
...
+              "hash_alg": "SHA256",
+              "hostkey": "ssh-rsa",
+            },
+            {
+              "hash": "a5:6f:62:26:81:03:b7:5e:06:48:10:04:79:4b:ac:32",
+              "hash_alg": "MD5",
+              "hostkey": "ssh-rsa",
+            },
+          ],
+          "host": "dummy-ssh.demo-targets.svc",
```
J12934, line 62: Should be `hostname`
Also does the result file format contain the info on which port the ssh server is running on?
So that we'd have this here as well (especially interesting for ssh service not running on the default port)

J12934 (changes requested):
This is looking really good 🙌
Noticed a couple of issues with the finding format. (Comment for that are primarily in the parser test)
Also a couple of suggestion on how to refactor the parser to make the code a bit more functional and consistent with the other parser code that we have.

scanners/ssh-audit/parser/parser.test.js
sofi0071: We changed it from `hostname` to `host` because ssh-audit doesn't differentiate between IP-address and hostname. So to avoid having `hostname: 127.0...` or `ip-address: dummy-ssh`, we put them together and named it host. Is there a better alternative?

scanners/ssh-audit/parser/parser.js
sofi0071: Yes it was a typo, thank you!

scanners/ssh-audit/parser/parser.js
sofi0071: Thank you, it calls Object.values twice because of the way the object is built. We have to extract the needed information twice.

scanners/ssh-audit/parser/parser.test.js
sofi0071: We decided we will name it `hostname` to avoid confusion.

J12934: Mh ok understandable, at least from a usage / consistency perspective I'd prefer to have both a hostname and a ip_address field and have one of them be null depending on what ssh-audit gives us.

This way you can more easily filter in Kibana for finding with `attributes.hostname` and get all results (which have a hostname) and you wouldn't need to search both in host and hostname. If somebody wants a straightforward field which has the location set they should use the `location` field.

sofi0071: okay, that makes sense. We changed it accordingly and now have hostname and ip_address.

J12934 (approved):

PR #1722, opened 2023-05-12
Added a crontab configuration option to scheduledScans

### LINKED ISSUE #150: ➹ Add a crontab configuration option to scheduledScans
**Is your feature request related to a problem? Please describe.**
As a security tester i would like to configure the start time pattern for [scheduledScans](https://docs.securecodebox.io/docs/api/crds/scheduled-scan) more precisely. For now it is possible to configure the time interval (hours) between to scans but the starting point relates to the installation time of the scheduledScan.

To be able to configure a more complex schedule for my security scan it would be great to support a [`crontab` syntax ](https://en.wikipedia.org/wiki/Cron) as an alternative option. 

#### Example
```yaml
kind: ScheduledScan
metadata:
  name: "nmap-scanme.nmap.org-daily"
spec:
  crontab: "0 12 * * *"
  scanSpec:
    scanType: "nmap"
    parameters:
        # Use nmaps service detection feature
        - "-sV"
        - scanme.nmap.org
  historyLimit: 3
```

none

**Additional context**
- About Cron: https://en.wikipedia.org/wiki/Cron
- GO Library for parsing https://godoc.org/github.com/robfig/cron
- https://docs.securecodebox.io/docs/api/crds/scheduled-scan

Ilyesbdlala:
## Description

closes #150 
Current status: Cronjob configuration works  ~~(tested manually), However no unit test have been implemented.~~ Unit tests are implemented that use polling to test if the cronjob has started. This is slower by a maximum of 60 seconds and a minimum of 0 seconds. the slow test has the golang build tag "slow" and other/fast tests have the tag "fast". 
Calling `make test` runs all tests i.e "slow and fast" 
Calling `make test-fast` runs only the tests with "fast" tag.

### Checklist

### DIFF
--- a/operator/controllers/execution/scheduledscan_controller.go
+++ b/operator/controllers/execution/scheduledscan_controller.go
@@ -171,7 +171,7 @@ func (r *ScheduledScanReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		}
 
 		// Recalculate next schedule
-		nextSchedule = r.Clock.Now().Add(scheduledScan.Spec.Interval.Duration)
+		nextSchedule, err = getNextSchedule(r, scheduledScan, r.Clock.Now())
 	}
 
 	return ctrl.Result{RequeueAfter: nextSchedule.Sub(r.Clock.Now())}, nil
@@ -263,7 +263,7 @@ func (r *ScheduledScanReconciler) deleteOldScans(scans []executionv1.Scan, maxCo
 func (r *ScheduledScanReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	// set up a real clock, since we're not in a test
 	if r.Clock == nil {
-		r.Clock = realClock{}
+		r.Clock = &realClock{}
 	}
 	ctx := context.Background()
 	if err := mgr.GetFieldIndexer().IndexField(ctx, &executionv1.Scan{}, ownerKey, func(rawObj client.Object) []string {
--- a/operator/controllers/execution/scheduledscan_controller.go
+++ b/operator/controllers/execution/scheduledscan_controller.go
@@ -17,6 +17,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -32,8 +33,9 @@ var (
 // ScheduledScanReconciler reconciles a ScheduledScan object
 type ScheduledScanReconciler struct {
 	client.Client
-	Log    logr.Logger
-	Scheme *runtime.Scheme
+	Log      logr.Logger
+	Scheme   *runtime.Scheme
+	Recorder record.EventRecorder
 }
 
 // +kubebuilder:rbac:groups=execution.securecodebox.io,resources=scheduledscans,verbs=get;list;watch;create;update;patch;delete
@@ -167,6 +169,7 @@ func getNextSchedule(r *ScheduledScanReconciler, scheduledScan executionv1.Sched
 	if scheduledScan.Spec.Schedule != "" {
 		sched, err := cron.ParseStandard(scheduledScan.Spec.Schedule)
 		if err != nil {
+			r.Recorder.Event(&scheduledScan, "Warning", "ScheduleParseError", fmt.Sprintf("Unparseable schedule %q: %v", scheduledScan.Spec.Schedule, err))
 			return time.Time{}, fmt.Errorf("Unparseable schedule %q: %v", scheduledScan.Spec.Schedule, err)
 		}
 
@@ -193,6 +196,7 @@ func getNextSchedule(r *ScheduledScanReconciler, scheduledScan executionv1.Sched
 		}
 		return nextSchedule, nil
 	}
+	r.Recorder.Event(&scheduledScan, "Warning", "NoScheduleOrInterval", "No valid schedule or interval found")
 	return time.Time{}, fmt.Errorf("No schedule or interval found")
 }
 
--- a/operator/controllers/execution/scheduledscan_controller.go
+++ b/operator/controllers/execution/scheduledscan_controller.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	"github.com/robfig/cron"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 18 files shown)


operator/controllers/execution/scheduledscan_controller_test.go
```diff
@@ -1,6 +1,10 @@
 // SPDX-FileCopyrightText: the secureCodeBox authors
 //
 // SPDX-License-Identifier: Apache-2.0
+
+//go:build fast
+// +build fast
```
J12934, line 6: What does this do?

operator/controllers/execution/suite_test.go
```diff
@@ -36,6 +37,17 @@ var testEnv *envtest.Environment
...
+	timeToAdd time.Duration
+}
+
+var FakeTime = time.Date(2023, 1, 1, 15, 0, 0, 0, time.UTC)
+var FakeClock = &fakeClock{timeToAdd: 0}
+
+func (f *fakeClock) Now() time.Time             { return FakeTime.Add(f.timeToAdd) }
+func (f *fakeClock) TimeTravel(d time.Duration) { f.timeToAdd += d }
+func (f *fakeClock) Reset()                     { f.timeToAdd = 0 }
+
```
J12934, line 50: Is this still used?

operator/controllers/execution/scheduledscan_controller.go
```diff
@@ -156,12 +156,46 @@ func (r *ScheduledScanReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+		var nextSchedule time.Time
+		if scheduledScan.Status.LastScheduleTime != nil {
+			nextSchedule = scheduledScan.Status.LastScheduleTime.Add(scheduledScan.Spec.Interval.Duration)
+		} else {
+			nextSchedule = time.Now().Add(-1 * time.Second)
+		}
+		return nextSchedule, nil
+	}
+	return time.Time{}, fmt.Errorf("No schedule or interval found")
+}
```
J12934, line 197: Do we have a (easy) way to validate that the user set either the cron or the interval?
Or at least add a error event to the scheduled scan to indicate to users that the config is invalid without them having to look through the operator logs?

operator/controllers/execution/scheduledscan_controller_test.go
Ilyesbdlala: Unit tests are implemented that use polling to test if the cronjob has started. This is slower by a maximum of 60 seconds and a minimum of 0 seconds. the slow test has the golang build tag "slow" and other/fast tests have the tag "fast".
Calling make test runs all tests i.e "slow and fast"
Calling make test-fast runs only the tests with "fast" tag.

operator/controllers/execution/suite_test.go
Ilyesbdlala: Not any more, I did not know at the time which solution we're going to use, so i kept both. I'll remove the fakeClock :)

operator/controllers/execution/scheduledscan_controller.go
Ilyesbdlala: I can turn this into 
```
if (Schedule ) 
...
else if (Interval)
... 
else 
return time.Time{}, fmt.Errorf("No schedule or interval found")
```
Would that be good enough ?

J12934: Mhhh isn't that basically what we have now?
The problem is that for users without access to the logs of the operator these kinds of errors would be completly invisible. Ideally these invalid ScheduledScans could not even get created. If that's to hard (because it'd probably require a custom validating webhook we could at least add attach a error event to the ScheduledScan so that users using `kubectl describe scheduledscan <name>` could at least see a hint whats wrong with it.

J12934: We do that for Scans where the scan type doesn't exist (as far as i remember)

Ilyesbdlala: Just pushed a possible implementation. I am running however into a problem. The events are not getting recorded due to permission issues. I get the following error log:
`
│ E0714 10:01:20.230395       1 event.go:267] Server rejected event '&v1.Event{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ObjectMeta:v1.ObjectMeta{Name:"old-typo3.demo-targets.svc.org-1min.1771b311efbe8e98", GenerateName:"", Namespace:"default", SelfLink:"", UID:"", ResourceVersion:"", Generation:0, CreationTimestamp:time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC), DeletionTimestamp:<nil>, DeletionGracePeriodSeconds:(*int64)(nil), Labels:map[string]string(nil), Annotations:map[string]string(nil), OwnerReferences:[]v1.OwnerReference(nil), Finalizers:[]string(nil), ZZZ_DeprecatedClusterName:"", ManagedFields:[]v1.ManagedFieldsEntry(nil)}, InvolvedObject:v1.ObjectReference{Kind:"ScheduledScan", Namespace:"default", Name:"old-typo3.demo-targets.svc.org-1min", UID:"e17ce9b3-4b7c-4d92-8157-3cea39efd8cb", APIVersion:"execution.securecodebox.io/v1", ResourceVersion:"1642", FieldPath:""}, Reason:"ScheduleParseError", Message:"Unparseable schedule \"/5 * * * *\": Failed to parse int from : strconv.Atoi: parsing \"\": invalid syntax", Source:v1.EventSource{Component:"ScheduledScanController", Host:""}, FirstTimestamp:time.Date(2023, time.July, 14, 9, 50, 24, 858705560, time.Local), LastTimestamp:time.Date(2023, time.July, 14, 10, 1, 20, 228309484, time.Local), Count:18, Type:"Warning", EventTime:time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC), Series:(*v1.EventSeries)(nil), Action:"", Related:(*v1.ObjectReference)(nil), ReportingController:"", ReportingInstance:""}': 'events "old-typo3.demo-targets.svc.org-1min.1771b311efbe8e98" is forbidden: User "system:serviceaccount:securecodebox-system:securecodebox-operator" cannot patch resource "events" in API group "" in the namespace "default"' (will not retry!) 
`


TLDR:
`
User "system:serviceaccount:securecodebox-system:securecodebox-operator" cannot patch resource "events" in API group "" in the namespace "default"' (will not retry!)                                 
`

I am wondering why are the permissions missing here (or not functioning correctly) but supposedly work here:
https://github.com/secureCodeBox/secureCodeBox/blob/main/operator/controllers/execution/scantype_controller.go#L87
Any idea?

operator/controllers/execution/scheduledscan_controller_test.go
J12934: Ah cool didn't know about that 🙌

J12934 (approved):

operator/controllers/execution/scheduledscan_controller.go
Ilyesbdlala: Reason: https://github.com/secureCodeBox/secureCodeBox/pull/1811

PR #1820, opened 2023-07-20
Fix failing build of DefectDojo persistance hook

Weltraumschaf:


Zero3141:
We should wait until the compatible `1.0.0-SNAPSHOT` version of our defectdojo-client is available at sonatype central. Otherwise this build will not compile.

Weltraumschaf:
> We should wait until the compatible `1.0.0-SNAPSHOT` version of our defectdojo-client is available at sonatype central. Otherwise this build will not compile.

Is it not available yet?

Zero3141 (approved):

PR #1854, opened 2023-08-09
Add a basic SBOM workflow (closes #1838)

### LINKED ISSUE #1838: Generate SBOMs

Currently the secureCodeBox serves as a security scanner and always generates findings. The same architecture can be leveraged to generate software bills of materials (SBOMs), to catalogue the software and components used. This is especially useful when combined with the auto discovery to automatically create an inventory of the running infrastructure.

The secureCodeBox should be able to generate SBOMs, initially for containers.

Tests and comparisons have shown that [trivy](https://github.com/aquasecurity/trivy) and [syft](https://github.com/anchore/syft) are the best tools to generate SBOMs for containers (most feature complete, easy to use, formats supported, components detected).

Since trivy is already integrated as a scanner, a first proof of concept can use trivy to generate SBOMs using the CycloneDX format, because then they can be sent to [Dependency-Track](https://github.com/DependencyTrack/dependency-track) using a hook. Support for SPDX and xml formats can be added later if needed.

Ideally SBOMs are created at build-time when the information available is the most accurate. For most languages detecting dependenies in a container is accurate enough though. The environment the secureCodeBox is used in also usually does not have access to the software at build-time.

Formats (SPDX or xml) could be flexibly configured, this can be added later.

### Additional context

There are some issues regarding trivy SBOMs and vulnerability detection in Dependency-Track, trivy only generates package urls (purls) for components, but Dependency-Track can only match issues in the NVD with common platform enumerations (CPEs). Issues not listed in other sources (OSS Index or the GitHub Advisory Database) will then not be found.

Syft generates CPEs, so it might make sense to also integrate syft in the future.

o1oo11oo:
## Description

This is an MVP for an SBOM workflow as part of the secureCodeBox. It adds
- a trivy-sbom scanner to generate SBOMs for container images
- a CycloneDX parser to generate findings
- a persistence hook to send CycloneDX SBOMs to Dependency-Track

There are some challenges regarding SBOM integrations. ~Since trivy can now generate outputs in a different format (CycloneDX SBOMs instead of trivy-json), the parser needs to be able to handle this. It is cleaner and more future-proof to create a separate CycloneD SBOM parser. This needs to be integrated and built somehow, currently it is part of the trivy scanner directory, because only trivy creates SBOMs.~ _Because of way too much trouble with one scanner with two parsers, trivy-sbom is now an individual scanner. The parser is a generic CycloneDX parser though._ If Syft is also integrated in the future, we might need to change the architecture to have global parsers which can be used for multiple scanners. ~The parser integration requires some Makefile changes, which could be nicer, but work for the time being.~

SBOM-related items and configurations are named generically without trivy so that moving them somewhere else or independently using them in the future is easier. I'm unsure though if SBOM or CycloneDX is a better name, currently only CycloneDX is supported and used, but in the future we might want to generate SPDX SBOMs as well, and then we need to decide if that should be the same scanType and parser, or if that should remain completely separated.

Hooks are triggered for all scans, but the Depenendency-Track hook is only useful for scans which create CycloneDX SBOMs, all other formats or findings cannot be handled. To mitigate that, the hook first checks the scan results type for `sbom-cyclonedx` and then checks that the raw results has the `bomFormat` property set to `CycloneDX`, a mandatory part of CycloneDX-SBOMs. All other scans are ignored.

The Dependency-Track hook automatically creates projects for SBOMs it uploads. For this the API requires a project name and a version, which the hook tries to get from the `name` property of the main component in the CycloneDX SBOM. This works as long as only image SBOMs are used, which have a name of `<repository>:<tag>`, where the hook splits by `:` and gets the first component as name, the last as version or defaults to `latest`. This probably stops working in the future when SBOMs for other targets are integrated as well, but CycloneDX SBOMs do not provide better properties to get this data, trivy SBOMs do not contain a version for the main component and syft only sets the sha256 hash.

Currently unimplemented features that we might still want at some point:
- ~tests for the parser-cyclonedx~ _unit tests and integration tests exist_
- ~tests for the persistence-dependencytrack hook~ _unit tests exist_
- reading vulnerabilities from the SBOM and generating findings from them

⚠️ This adds a new Docker image/repository `securecodebox/parser-cyclonedx`, which needs to be registered on Docker Hub first, otherwise the next release will fail.

Closes #1838.

### Checklist


hooks/persistence-dependencytrack/Chart.yaml
```diff
@@ -0,0 +1,30 @@
...
+name: persistence-dependencytrack
+description: The dependencytrack persistence provider persists secureCodeBox CycloneDX SBOMs into Dependency-Track.
+
+type: application
+
+# version - gets automatically set to the secureCodeBox release version when the helm charts gets published
+version: "4.0.0"
+
+# TODO does it make sense to set this? DefectDojo's seems quite outdated, docs say set this and version to latest
+appVersion: "4.8.2"
```
o1oo11oo, line 15: The docs said to set this value, but I'm not sure how much sense it makes, I compared it to the DefectDojo hook and the value there is quite outdated. Probably it just needs to be here? This is the current Dependency-Track version that I tested against, but not the version of the hook.

o1oo11oo:
I went through the `Makefile` changes with @Ilyesbdlala and we changed the `deploy` invocation to prevent helm getting called twice. It's still not a perfect integration but good enough for an MVP.

scanners/trivy/templates/trivy-scan-type.yaml
```diff
@@ -297,4 +297,54 @@ spec:
...
             {{- toYaml .Values.scanner.extraVolumes | nindent 12 }}
-
+---
+apiVersion: "execution.securecodebox.io/v1"
+kind: ScanType
+metadata:
+  name: "trivy-image-sbom{{ .Values.scanner.nameAppend | default ""}}"
+spec:
+  extractResults:
+    type: sbom-cyclonedx
```
o1oo11oo, line 307: One possible name inconsistency is here, the scan is named `trivy-image-sbom`, because generating an SBOM in SPDX format is a simple parameter change to the trivy command (`--format spdx-json`), but the output format is named `sbom-cyclonedx` because the parser can (currently) not handle SPDX and neither can the Dependency-Track hook.

So this is a bit of a design question, how far separated do we want to handle SPDX SBOMs in case we add them as well? With completely separated formats/parsers, or integrated under the SBOM name?

J12934: I think the names for the helm chart "trivy-sbom" and for the result type "sbom-cyclondx" are good.
The scanType could maybe be changed to `trivy-image-sbom-cyclondx` if there comes a need to also support trivy spdx we can add it a second scanType there. (Though then we might get troubles again with having two parsers in one scanner)

I'm fine either way, we can leave it like it is and rename the scanType once / if we actually have a need to also support spdx.

J12934 (commented):
very nice work 👏
just some minor comments.

tested it out locally and worked perfectly on the first try :)

hooks/persistence-dependencytrack/hook/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "version": "1.0.0",
+  "description": "secureCodeBox hook to persist CycloneDX SBOMs to Dependency-Track.",
+  "homepage": "https://www.secureCodeBox.io",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/secureCodeBox/secureCodeBox.git"
+  },
+  "main": "hook.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
```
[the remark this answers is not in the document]
o1oo11oo: I amended this into the original "Add tests" commit

hooks/persistence-dependencytrack/Chart.yaml
o1oo11oo: I guess for keeping it similar to the other hooks I'll leave it in here for now, but I opened #1959 to keep track of this.

scanners/trivy/templates/trivy-scan-type.yaml
o1oo11oo: Well at this point the names are a bit different, because `trivy-sbom` is its own scanner, and I didn't want to mix up the names too much, so the scanType is called `trivy-sbom-image`, which I think sounds worse than `trivy-image-sbom`, but follows our usual convention of naming scanTypes after `<scanner>-<target>`.

But if you're otherwise fine with that I'd just keep the names like they are for now. Having one SBOM "parser", that can parse both CycloneDX and SPDX SBOMs also doesn't sound too bad to me, you could detect which one it is and then switch in the beginning. Much more than generating an informational finding "SBOM created" doesn't happen anyway. So instead of completely seperating the code probably only the information-extraction functions to fill that one informational finding are going to differ.

hooks/persistence-dependencytrack/hook/hook.js
```diff
@@ -0,0 +1,63 @@
+// SPDX-FileCopyrightText: the secureCodeBox authors
```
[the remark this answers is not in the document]
o1oo11oo: Thanks to the node 18 upgrade :)

hooks/persistence-dependencytrack/.helm-docs.gotmpl
```diff
@@ -0,0 +1,50 @@
...
+
+{{- define "extra.dockerDeploymentSection" -}}
+## Supported Tags
+- `latest`  (represents the latest stable release build)
+- tagged releases, e.g. `3.0.0`, `2.9.0`, `2.8.0`, `2.7.0`
+{{- end }}
+
+{{- define "extra.chartAboutSection" -}}
+## What is "Persistence Dependency-Track" Hook about?
+The Dependency-Track persistenceProvider hook saves all generated CycloneDX SBOMs into the configured [OWASP Dependency-Track][dependencytrack.org] instance, other findings or SPDX SBOMs cannot be handled and are ignored. This allows automatically cataloging infrastructure to gain an overview over the used components and dependencies. To learn more about Dependency-Track visit [dependencytrack.org].
```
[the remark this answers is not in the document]
o1oo11oo: I added links in both directions, from the scanner to the hook and the other way around. I realized though that using relative links can lead to problems with the generated READMEs and created #1963 for that.

hooks/persistence-dependencytrack/hook/hook.js
```diff
@@ -0,0 +1,63 @@
...
+    headers: {
+      "X-API-Key": apiKey,
+    },
+    body: formData,
+  });
+
+  // Response-token can be used to determine if any task is being performed on the BOM
+  // Endpoint: <url>/api/v1/bom/<token>
+  const content = await response.json();
+  console.log(`Successfully uploaded SBOM to Dependency-Track. Response-token: ${content.token}`);
```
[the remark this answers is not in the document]
o1oo11oo: Yeah makes sense, I added some error handling there.

scanners/trivy/templates/trivy-scan-type.yaml
o1oo11oo: @J12934  I guess renaming the scanType would be a breaking change? So maybe we should think about which name we want again, otherwise adding SPDX in the future might be a bit akward with scanType names `trivy-sbom-image` and `trivy-sbom-image-spdx` or something like that.

J12934 (approved):

J12934:
Look very good now 🚀
Tested another time and still working perfectly :)

scanners/trivy/templates/trivy-scan-type.yaml
o1oo11oo: Names work for now, can be merged

Issue #1852, opened 2023-08-07
Containers still use node:16


o1oo11oo:
### Describe the bug

Our containers are still using node:16 even though node:18 is the current LTS.

### Expected behavior

We should probably switch to a more recent version.

PR #1896, opened 2023-09-05
Upgrade node to latest LTS in templates

o1oo11oo:
I missed this node version when upgrading to latest LTS in #1853

Relates to #1852

### Checklist


Ilyesbdlala:
As far as I can tell that variable is never used (and is not needed) right ? So we can probably just remove it.

o1oo11oo:
Good point, I'm not sure, I didn't check the workflow that far. The other versions are also quite old (usually we use Go 1.19 I think).

o1oo11oo:
I opened #1946 to remove the unused version varaibles, so this now only contains the change of the node version in the templates.

Ilyesbdlala (approved):

Issue #942, opened 2022-01-26
Monorepo


Weltraumschaf:
It would make things easier to move the documentation into the main repo.

Weltraumschaf:
Open questions:

- [ ] do we need `release.md` anymore because we automated the release?
- [ ] do we need `developer-guide.md` anymore because we have this maybe better documented in the main documentation?
- [ ] do we need `user-guide.md` anymore because we have this maybe better documented in the main documentation?
- [ ] do we need `compononents_overview.uxf` anymore?
- [ ] do we need `sequence_overview.puml` anymore because we have https://www.securecodebox.io/docs/architecture/runtime_view?

Weltraumschaf:
This bug relates to this: https://github.com/secureCodeBox/secureCodeBox/issues/1963

PR #1944, opened 2023-09-11
Remove duplicate sponsor images from documentation

Zero3141:
## Description
Part of [#942](https://github.com/secureCodeBox/secureCodeBox/issues/942)

### Checklist


documentation/src/utils/config.js
```diff
@@ -4,7 +4,7 @@
 
 const docsConfig = {
     repository: "secureCodeBox/secureCodeBox", // The repository url without the github part of the link
-    targetPath: "docs", // This needs to be 'docs' for the docusaurus build, but you may specify a 'docs/<subdirectory>'
+    targetPath: ".", // This needs to be '.' as files copied to docs (scanners & hooks) and static (sponsors) folder
```
Weltraumschaf, line 7: Id do not really understand the comment here.

documentation/src/utils/config.js
```diff
@@ -28,12 +28,16 @@ const docsConfig = {
...
-        dst: "scanners",
+        dst: "docs/scanners",
       },
       {
         src: "hooks",
-        dst: "hooks",
+        dst: "docs/hooks",
       },
+      {
+        src: "resources/sponsors",
```
Weltraumschaf, line 38: Why that? We can simply reference the images IMO in the main README directly, right?

Weltraumschaf (changes requested):
We should remove any unnecessary code to build things. So IMO the images should reside in the docusaurus folder, so we do not need to move them around. In the main README we can simply deep link to these files.

Also we should consider if we can simplify the scanner/hooks docs generation, to get rid ofthe untested build scripts.

Zero3141:
Replaced by https://github.com/secureCodeBox/secureCodeBox/pull/1950

PR #2133, opened 2023-12-14
Add a wait before refetching DefectDojo results (closes #2057)

### LINKED ISSUE #2057: DefectDojo findings are not synced back correctly
### Describe the bug

I am seeing issues with my synced back findings.

My setup includes DefectDojo, the DefectDojo persistence provider, the notification hook and the ZAP + Nikto scanners.

I am trying to prevent a notification from being sent in case all findings in are already "handled" (mitigated as false positive or risk accepted) in DefectDojo as described shortly here: https://www.securecodebox.io/docs/hooks/notification-webhook/#configuration-of-a-notification

I have three findings in DefectDojo which were found by ZAP (Automation Framework) and imported using the persistence hook.

After doing a risk acceptance I expect all of them to be marked as duplicates during the next scan and all should include the "defectdojo.org/original-finding" data.

However as you can see in the json attached the last finding is not marked as a duplicate and the original finding is not attached.

[findings-zap.json](https://github.com/secureCodeBox/secureCodeBox/files/13053912/findings-zap.json)

With Nikto I am seeing the same issues, but with two (out of five) findings not have the details included. From the picture below the Alt-SVC finding was correctly duplicated and the X-Content finding was not.

Interestingly they are also the last ones in the json! Is there any sorting going on or is it just a coincidence?

### Steps To Reproduce
1. Create DefectDojo Persistence Setup as described below
2. Run scans and import the into DefectDojo through the provider
3. Check the synced back to json for the missing details.

### Expected behavior
I expect the synced back findings to be marked as a duplicate and to include the original DefectDojo finding as described here

 - secureCodeBox Version/Release: 4.1.0
 - Kubernetes Version: v1.24.14-gke.2700

### Screenshots / Logs

**ZAP Findings in DefectDojo**
image

**Nikto Findings in DefectDojo**
image

**Persistence Hook Log**
```
 @LT100144 ~ k logs securecodebox-defectdojo-persistence-defectdojo-nikto-sch-rws67
2023-10-20 11:34:21 INFO  DefectDojoPersistenceProvider:25 - Starting DefectDojo persistence provider
2023-10-20 11:34:21 INFO  DefectDojoPersistenceProvider:36 - Downloading Scan Result
2023-10-20 11:34:22 INFO  DefectDojoPersistenceProvider:40 - Uploading Findings to DefectDojo at: https://REDACTED
2023-10-20 11:34:22 INFO  VersionedEngagementsStrategy:85 - Running with DefectDojo User Id: 12
2023-10-20 11:34:22 INFO  VersionedEngagementsStrategy:203 - Looking for ID of ProductType 'Customer Facing Infrastructure'
2023-10-20 11:34:22 INFO  VersionedEngagementsStrategy:210 - Using ProductType Id: 7
2023-10-20 11:34:24 INFO  VersionedEngagementsStrategy:115 - Uploaded Scan Report as testID 2668 to DefectDojo
2023-10-20 11:34:24 INFO  DefectDojoPersistenceProvider:44 - Identified total Number of findings in DefectDojo: 5
2023-10-20 11:34:24 INFO  DefectDojoPersistenceProvider:51 - Overwriting secureCodeBox findings with the findings from DefectDojo.
2023-10-20 11:34:25 INFO  S3Service:25 - Uploading Findings to S3
2023-10-20 11:34:26 INFO  KubernetesService:88 - Updating Scan metadata
2023-10-20 11:34:26 INFO  DefectDojoPersistenceProvider:63 - DefectDojo Persistence Completed
```

### Additional context

**Helm Setup**
```
resource "helm_release" "securecodebox-defectdojo" {
  name       = "securecodebox-defectdojo"
  repository = "https://charts.securecodebox.io"
  chart      = "persistence-defectdojo"
  namespace  = kubernetes_namespace_v1.securecodebox.metadata[0].name
  version    = "4.1.0"

  values = [
    <<-EOF
    hook:
      tolerations:
      - key: "instance_type"
        value: "preemptible"
        effect: "NoSchedule"
    EOF
  ]

  set {
    name = "defectdojo.url"
    value = "https://REDACTED"
  }

  set {
    name = "defectdojo.authentication.userSecret"
    value = kubernetes_secret_v1.defectdojo-scanner-user-credentials.metadata[0].name
  }

  set {
    name = "hook.resources.limits.memory"
    value = "500Mi"
  }
}

resource "helm_release" "securecodebox-nikto" {
  name       = "securecodebox-nikto"
  repository = "https://charts.securecodebox.io"
  chart      = "nikto"
  namespace  = kubernetes_namespace_v1.securecodebox.metadata[0].name
  version    = "4.0.1"

  values = [
    <<-EOF
    parser:
      tolerations:
      - key: "instance_type"
        value: "preemptible"
        effect: "NoSchedule"
    scanner:
      tolerations:
      - key: "instance_type"
        value: "preemptible"
        effect: "NoSchedule"
    EOF
  ]
}

resource "helm_release" "securecodebox-zap" {
  name       = "securecodebox-zap"
  repository = "https://charts.securecodebox.io"
  chart      = "zap"
  namespace  = kubernetes_namespace_v1.securecodebox.metadata[0].name
  version    = "4.0.1"

  values = [
    <<-EOF
    parser:
      tolerations:
      - key: "instance_type"
        value: "preemptible"
        effect: "NoSchedule"
    scanner:
      tolerations:
      - key: "instance_type"
        value: "preemptible"
        effect: "NoSchedule"
    EOF
  ]

  set {
    name = "scanner.resources.requests.cpu"
    value = "2200m"
  }

  set {
    name = "scanner.resources.requests.memory"
    value = "8200Mi"
  }

  set {
    name = "scanner.resources.limits.memory"
    value = "8200Mi"
  }
}

}

resource "helm_release" "securecodebox-notification" {
  name       = "securecodebox-notification"
  repository = "https://charts.securecodebox.io"
  chart      = "notification"
  namespace  = kubernetes_namespace_v1.securecodebox.metadata[0].name
  version    = "4.1.0"

  values = [
    <<-EOF
    customTemplateMap:
      exists: true
      name: "${kubernetes_config_map_v1.email-template.metadata[0].name}"
    hook:
      tolerations:
      - key: "instance_type"
        value: "preemptible"
        effect: "NoSchedule"
    notificationChannels:
      - name: email
        type: email
        template: REDACTED
        skipNotificationOnZeroFinding: true
        rules:
        - matches:
            anyOf:
            - category: "DefectDojo Imported Finding"
              attributes:
                defectdojo.org/original-finding:
                  riskAccepted: false
            - category: "DefectDojo Imported Finding"
              attributes:
                duplicate: false
        endPoint: "REDACTED"
    env:
      - name: EMAIL_FROM
        value: REDACTED
      - name: SMTP_CONFIG
        valueFrom:
          secretKeyRef:
            name: email-config
            key: smtp-config
    EOF
  ]
}


```

moxli:
## Description
My aim is to resolve issue #2057 with this change by adding a wait before the refetching the DefectDojo results.

### Checklist

### DIFF
--- a/hooks/persistence-defectdojo/hook/src/main/java/io/securecodebox/persistence/service/DeduplicationAwaitingService.java
+++ b/hooks/persistence-defectdojo/hook/src/main/java/io/securecodebox/persistence/service/DeduplicationAwaitingService.java
@@ -1,0 +1,33 @@
+// SPDX-FileCopyrightText: the secureCodeBox authors
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package io.securecodebox.persistence.service;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.TimeUnit;
+
+public class DeduplicationAwaitingService {
+    private static final Logger LOG = LoggerFactory.getLogger(DeduplicationAwaitingService.class);
+
+    Integer refetchWaitSeconds;
+
+    public void refetchWait() {
+      try {
+        refetchWaitSeconds = Integer.valueOf(System.getenv("DEFECTDOJO_REFETCH_WAIT_SECONDS"));
+        if (refetchWaitSeconds > 0) {
+          LOG.info("Waiting for {} seconds for deduplication to finish before continuing", refetchWaitSeconds);
+          try {
+            TimeUnit.SECONDS.sleep(refetchWaitSeconds);
+          } catch (InterruptedException e) {
+            LOG.warn("Sleeping failed: ", e);
+          }
+        }
+      } catch (NumberFormatException e) {
+          LOG.warn("Could not convert the value of DEFECTDOJO_REFETCH_WAIT_SECONDS to an integer: ",e);
+      }
+    }
+}
+ 
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -51,3 +51,4 @@ Committing with `git commit -s` will add the sign-off at the end of the commit m
 - Lukas Fischer <lukas.fischer@iteratec.com>
 - Heiko Kiesel <heiko.kiesel@iteratec.com>
 - Frank Belter <frank.belter@iteratec.com>
+- Maximilian Dorner <maximilian.dorner@clever-soft.com>
--- a/hooks/persistence-defectdojo/hook/src/main/java/io/securecodebox/persistence/strategies/VersionedEngagementsStrategy.java
+++ b/hooks/persistence-defectdojo/hook/src/main/java/io/securecodebox/persistence/strategies/VersionedEngagementsStrategy.java
@@ -13,6 +13,7 @@ import io.securecodebox.persistence.defectdojo.model.*;
 import io.securecodebox.persistence.defectdojo.service.*;
 import io.securecodebox.persistence.exceptions.DefectDojoPersistenceException;
 import io.securecodebox.persistence.models.Scan;
+import io.securecodebox.persistence.service.DeduplicationAwaitingService;
 import io.securecodebox.persistence.util.DescriptionGenerator;
 import io.securecodebox.persistence.util.ScanNameMapping;
 import org.apache.commons.collections4.map.LinkedMap;
@@ -25,7 +26,6 @@ import java.time.format.DateTimeFormatter;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.concurrent.TimeUnit;
 
 /**
  * VersionedEngagementsStrategy creates a new Engagement for every new version of the software.
@@ -47,6 +47,7 @@ public class VersionedEngagementsStrategy implements Strategy {
   TestTypeService testTypeService;
   ImportScanService importScanService;
   FindingService findingService;
+  DeduplicationAwaitingService deduplicationAwaitingService;
 
   Config config;
   PersistenceProviderConfig persistenceProviderConfig;
@@ -66,6 +67,7 @@ public class VersionedEngagementsStrategy implements Strategy {
     this.testTypeService = new TestTypeService(defectDojoConfig);
     this.importScanService = ImportScanService.createDefault(defectDojoConfig);
     this.findingService = new FindingService(defectDojoConfig);
+    this.deduplicationAwaitingService = new DeduplicationAwaitingService();
 
     this.config = defectDojoConfig;
     this.persistenceProviderConfig = persistenceProviderConfig;
@@ -115,12 +117,8 @@ public class VersionedEngagementsStrategy implements Strategy {
 
     LOG.info("Uploaded Scan Report as testID {} to DefectDojo", testId);
 
-    var refetchWaitSeconds = config.getRefetchWaitSeconds();
-    if (!(refetchWaitSeconds == 0)) {
-      LOG.info("Waiting for {} seconds before continuing...", refetchWaitSeconds);
-      TimeUnit.SECONDS.sleep(refetchWaitSeconds);
-    }
-
+    deduplicationAwaitingService.refetchWait();
+    
     return findingService.search(Map.of("test", String.valueOf(testId)));
   }
 
--- a/hooks/persistence-defectdojo/hook/src/main/java/io/securecodebox/persistence/strategies/VersionedEngagementsStrategy.java
+++ b/hooks/persistence-defectdojo/hook/src/main/java/io/securecodebox/persistence/strategies/VersionedEngagementsStrategy.java
@@ -25,6 +25,7 @@ import java.time.format.DateTimeFormatter;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.TimeUnit;
 
 /**
  * VersionedEngagementsStrategy creates a new Engagement for every new version of the software.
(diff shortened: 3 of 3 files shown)


moxli:
I would love to get some feedback on the best approach for this change.
I am worried about the best location for the wait/sleep.

I would prefer to have it within the following if-clause as a separate step because I feel like there it makes the most sense: https://github.com/secureCodeBox/secureCodeBox/blob/main/hooks/persistence-defectdojo/hook/src/main/java/io/securecodebox/persistence/DefectDojoPersistenceProvider.java#L46

However at that point the defectDojoFindings variable is populated already, so a sleep won't change the outcome.

That fastest (arguably hacky) way I came up with, was to put it into VersionedEngagementsStrategy.

It might be better to change the way the results are fetched? Or do a refetch in case the hook is ReadWrite?

The way I tested it without refetching the the findings I had to add the parameter to the defectdojo java client.
The changes are described here:
https://github.com/secureCodeBox/defectdojo-client-java/issues/98
https://github.com/secureCodeBox/defectdojo-client-java/pull/99

I also feel like implementing this feature completely within the hook itself makes more sense.

Please let me know your thoughts on this.
Thank you :)

J12934:
@moxli awesome 🚀
I think doing the wait from the VersionedEngagementsStrategy really is the best approach, assuming we don't want to change the interface of the Strategy class. 

The goal there was initially to have the ability to write different strategies which upload the findings in DefectDojo in totally different ways. So far the approach taken by the VersionedEngagementsStrategy has been enough for everyone so we never bothered writing another one. But as the goal of the strategy was to support widely different approaches of using DD we made the interface as generic as it can be. So it needs to already return the findings, as depending on the approach/strategy) they might need to be refetched differently. So every Strategy would need to implement the waiting for deduplication to finish in some way on their own.

Maybe to make this more "generic" / cleaner this would be put in something like a `DeduplicationAwaitingService` so that it could be reused in case we ever need a second strategy.

moxli:
@J12934 okay that sounds great :) 

Should the DeduplicationAwaitingService be part of the hook or the java client? 

I would love to have the logging for it be a part of the service and it would be easier to simply use the Logger of the KubernetesService if the it was part of the hook.

Or maybe use return the number of seconds it got from the env var and log it in the hook?

Weltraumschaf:
Hi @moxli,

thanks for you're contribution. Since it is very close to x-mas, everyone is very busy organizing the last gifts and such 😬

Maybe we have time between x-mas and new year to look into this.

moxli:
@J12934 I have moved the wait logic to a separate service as discussed. :) 
Do you have an idea how this service could be tested? I am not that familiar with this tbh.

hooks/persistence-defectdojo/hook/src/main/java/io/securecodebox/persistence/service/DeduplicationAwaitingService.java
```diff
@@ -0,0 +1,33 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.TimeUnit;
+
+public class DeduplicationAwaitingService {
+    private static final Logger LOG = LoggerFactory.getLogger(DeduplicationAwaitingService.class);
+
+    Integer refetchWaitSeconds;
```
Weltraumschaf, line 15: For what reason is this a object property, since it is only used in the method `#refetchWait()`?

hooks/persistence-defectdojo/hook/src/main/java/io/securecodebox/persistence/service/DeduplicationAwaitingService.java
```diff
@@ -0,0 +1,33 @@
...
+
+    public void refetchWait() {
+      try {
+        refetchWaitSeconds = Integer.valueOf(System.getenv("DEFECTDOJO_REFETCH_WAIT_SECONDS"));
```
Weltraumschaf, line 19: It is good practice in the Java world to extract such a string into a constant (`private static final String`).

I just saw in the code that there are some other `System.getenv()` scattered in the hook's source. That's not good for some reasons:

1. `Syste.getenv()` is a side effect and makes unit testing hard.
2. Such constants and config gathering should be at one place.

I consider to provide a patch to address these two topics.

Weltraumschaf:
@J12934 Why is it called service?

IMO a good design would be: Extract the configuration gathering (all the `System#getenv()` into a separate class and name it just awaiter:

```java
 @Slf4j
    public final class EnvConfig {
        private static final String REFETCH_WAIT_SECONDS = "DEFECTDOJO_REFETCH_WAIT_SECONDS";

        public int refetchWaitSeconds() {
            final var raw = System.getenv(REFETCH_WAIT_SECONDS);

            try {
                return Math.max(0, Integer.parseInt(raw));
            } catch (NumberFormatException e) {
                log.warn("Could not convert the value {} of {} to an integer!", raw, REFETCH_WAIT_SECONDS);
                return 0;
            }
        }
    }

    @Slf4j
    public final class Awaiter {
        private final int secondsToWait;

        public Awaiter(int secondsToWait) {
            super();
            this.secondsToWait = secondsToWait;
        }

        public void await() {
            if (secondsToWait > 0) {
                try {
                    TimeUnit.SECONDS.sleep(secondsToWait);
                } catch (InterruptedException e) {
                    log.warn("Waiting failed!", e);
                }
            }
        }
    }
```
And usage:
```java
final var config = new EnvConfig();
// ...
final var awaiter = new Awaiter(config.refetchWaitSeconds());
// ...
log.info("Waiting for {} seconds for deduplication to finish before continuing", config.refetchWaitSeconds());
awaiter.await();
```

Then we can move all the other `System#getenv()` into `Envconfig`.

hooks/persistence-defectdojo/hook/src/main/java/io/securecodebox/persistence/service/DeduplicationAwaitingService.java
moxli: @Weltraumschaf I should probably have added a comment that I am by no means a Java developer ^^
I am happy, that I got it working at all.

What exactly should I change here?

Weltraumschaf: No problem, I jump in, i've some experience in Java :-)

Weltraumschaf:
I've implemented it in a new PR: https://github.com/secureCodeBox/secureCodeBox/pull/2208

PR #2177, opened 2024-01-12
Added Unit tests for CI helpers

Ilyesbdlala:
### Checklist

### DIFF
--- a/scanners/helpers.js
+++ b/scanners/helpers.js
@@ -130,7 +130,7 @@ async function disasterRecovery(scanName, CRDApi = getKubernetesAPIs().k8sCRDApi
  * @returns {scan.findings} returns findings { categories, severities, count }
  */
 async function scan(name, scanType, parameters = [], timeout = 180, volumes = [], volumeMounts = [],
-               initContainers = [], CRDApi = getKubernetesAPIs().k8sCRDApi, BatchApi = getKubernetesAPIs().k8sBatchApi, PodsApi = getKubernetesAPIs().k8sPodsApi) {
+  initContainers = [], CRDApi = getKubernetesAPIs().k8sCRDApi, BatchApi = getKubernetesAPIs().k8sBatchApi, PodsApi = getKubernetesAPIs().k8sPodsApi) {
   namespace = "integration-tests"
   const scanDefinition = {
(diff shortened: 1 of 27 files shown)


J12934:
is this already running / tested in the CI?

Ilyesbdlala:
> is this already running / tested in the CI?

Now it is :)

scanners/helpers.test.js
```diff
@@ -0,0 +1,193 @@
...
+          180,
+          mockK8sCRDApi,
+          mockK8sBatchApi,
+          mockPodsApi
+        );
+      } catch (error) {
+        expect(error).toMatchInlineSnapshot(
+          `[Error: Initial Scan failed with description "Mocked Error"]`
+        );
+      }
```
J12934, line 190: these kinds of assertions are a bit risky as the test would pass when there is not exception thrown.

there a two ways to avoid this in jest. you can either tell jest how many assertions to expect and fail if you have fewer.

or

(i prefer that way :) ): Use expect.rejects: https://jestjs.io/docs/expect#rejects
```js
test('rejects to octopus', async () => {
  await expect(Promise.reject(new Error('octopus'))).rejects.toThrow('octopus');
});
```

scanners/helpers.js
```diff
@@ -7,16 +7,27 @@ const k8s = require("@kubernetes/client-node");
...
+
+  return { k8sCRDApi, k8sBatchApi, k8sPodsApi };
+}
 let namespace = "integration-tests";
 
 const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms * 1000));
 
-async function deleteScan(name) {
-  await k8sCRDApi.deleteNamespacedCustomObject(
+async function deleteScan(name, CRDApi = getKubernetesAPIs().k8sCRDApi) {
```
J12934, line 29: I think I'd personally change these arguments in all of these functions to get in the entire getKubernetesAPIs as one bundle.

So basically every function would get a argument called k8sApis which then has the CrdApi, BatchApi, PodsApi on it. (Like it's already returned by the `getKubernetesAPIs()` function) Passing these as seperate arguments seems more complicated to me

J12934 (approved):

PR #2205, opened 2024-01-19
Fixed Scans being marked as Failed after the First Jobs has failed

### LINKED ISSUE #238: Scans are marked as Failed after the First Jobs has failed
## Current Status

The secureCodeBox Operator marks Scans as "Failed" if the job fails, even when the job gets restarted by Kubernetes and succeeds later.

## Expected Behavior

The scan is only marked as "Failed" if the retry limit for the job has been exceeded.

Ilyesbdlala:
## Description

Changed the condition for failed job to requiring exceding the specified BackoffLimit
BackoffLimit can be set in the values.yaml of scanners 
This allows that a scan  is only considered as Failed if all "retries" have failed
closes #238
### Checklist

### DIFF
--- a/operator/controllers/execution/scans/job.go
+++ b/operator/controllers/execution/scans/job.go
@@ -69,7 +69,7 @@ func (r *ScanReconciler) checkIfJobIsCompleted(scan *executionv1.Scan, labels cl
 
 	r.Log.V(9).Info("Got related jobs", "count", len(jobs.Items))
 
-	return checkIfAllJobsCompleted(jobs, r), nil
+	return checkIfAllJobsCompleted(jobs), nil
 }
 
 // injectCustomCACertsIfConfigured injects CA Certificates to /etc/ssl/certs/
--- a/operator/controllers/execution/scans/job.go
+++ b/operator/controllers/execution/scans/job.go
@@ -29,7 +29,7 @@ func checkIfAllJobsCompleted(jobs *batch.JobList) jobCompletionType {
 	hasCompleted := true
 
 	for _, job := range jobs.Items {
-		if job.Status.Failed > *job.Spec.BackoffLimit {
+		if job.Status.Failed >= *job.Spec.BackoffLimit {
 			return failed
 		} else if job.Status.Succeeded == 0 {
 			hasCompleted = false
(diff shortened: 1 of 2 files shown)


operator/controllers/execution/scans/job.go
```diff
@@ -29,7 +29,7 @@ func checkIfAllJobsCompleted(jobs *batch.JobList) jobCompletionType {
 	hasCompleted := true
 
 	for _, job := range jobs.Items {
-		if job.Status.Failed > 0 {
+		if job.Status.Failed > *job.Spec.BackoffLimit {
```
Weltraumschaf, line 32: What about input validation? What if one configurs "-42" in or "snafu" in the values.yaml?

Ilyesbdlala: Then it would throw an exception because of a problem between keyboard and chair. 
But I can add a default value, if job.Spec.BackoffLimit is not a positive number. That enough?

Weltraumschaf: 1. A default value makes sense, if the value is not configured or missed in values.yaml. Can this happen or is it this value somehow enforced?
2. Is the thrown exception to signal PEBCAK meaningful or is it a rather a FU-Exception which requires core-contributor skills to understand?

Ilyesbdlala: Also backoffLimit is defined a follows:
```go
	// Specifies the number of retries before marking this job failed.
	// Defaults to 6
	// +optional
	BackoffLimit *int32 `json:"backoffLimit,omitempty" protobuf:"varint,7,opt,name=backoffLimit"`
```
"snafu" would not accepted:
```bash
$ helm install nmap ./
Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(ScanType.spec.jobTemplate.spec.backoffLimit): invalid type for io.securecodebox.execution.v1.ScanType.spec.jobTemplate.spec.backoffLimit: got "string", expected "integer"
```
-42 is accepted but does not affect the logic, it behaves as if it was 0

Weltraumschaf: Ok, both should be mentioned in the commit message bc it is implicit knowledge and future Ilyes will be thankful for this info ;-)

Alon-Katz:
Hi, are there any updates on this PR? We are experiencing a serious issue that this PR could solve, and we would greatly appreciate it if we could proceed with it.

Ilyesbdlala:
> Hi, are there any updates on this PR? We are experiencing a serious issue that this PR could solve, and we would greatly appreciate it if we could proceed with it.
Working on it :) The tests just need to be updated, and the fix should be in the next release.

J12934 (approved):

PR #2348, opened 2024-03-08
Streamline CRD and RBAC Generation to avoid futher Issues

J12934:
## Description

This PR changes the controller-manager config to directly generate the CRDs and RBAC rules in the dirs used by the helm chart rather than the (mostly) unused config folders. These were autogenerated by kubebuilder but we stopped using these quite a while back. So I completly removed them to remove some hassle.

### Checklist


Ilyesbdlala:
Nice that we finally resolve this ^_^ Is the addition of the nodeSelector to the parsedefinition just something that was forgotten to be copied over to the `crd/` folder in another PR?

J12934:
> Nice that we finally resolve this ^_^ 
> 
> Is the addition of the nodeSelector to the parsedefinition just something that was forgotten to be copied over to the `crd/` folder in another PR?

Not sure if it was forgotten or later overwritten by something else. But it definitely should have been there 😅

PR #2539, opened 2024-06-26
Upgrade Node.js Versions to 22

J12934:
## Description

Upgrades node.js across the project to the latest lts 22

### Checklist


.github/workflows/ci.yaml
```diff
@@ -662,6 +601,11 @@ jobs:
           # - persistence-static-report (WIP)
     steps:
       - uses: actions/checkout@master
+
+      - uses: actions/setup-node@v4
+        with:
```
Weltraumschaf, line 606: Why here "with" instead of "node-version-file" used above?

J12934: hä? they are all:

I moved the node install up to the start of the action run here to make it consistent with the other jobs in the pipeline.

```yaml
      - uses: actions/setup-node@v4
        with:
          node-version-file: '.nvmrc'
```

Weltraumschaf: Ah, I see: In the others there are some more lines removed, so overlooked the "with".

Weltraumschaf (approved):

PR #2576, opened 2024-07-17
Add podSecurityContext to zap-scan-type template

eliihen:
## Description

The field `scanner.podSecurityContext` is documented as being supported by `zap-automation-framework`, but is not actually added to the template. This change adds it to the template so it is actually put onto the job when run.

### Checklist


Weltraumschaf:
Hi @eliihen,

thanks for submitting this issue. We'll investigate this.

J12934 (approved):
Oh good catch.
Thank you for the fix 🙌

Issue #189, opened 2020-11-02
secureCodeBox CLI (scbctl)


fuhrmeistery:
**Is your feature request related to a problem? Please describe.**

As user / developer of the *secureCodeBox* I would like to interact with the *secureCodeBox* via an unified CLI to fulfill tasks such as installing, start, update and delete scanners / hooks, generate stubs for new scanners / hooks

**Describe the solution you'd like**

There is a CLI tool to: 

- [x] create & observe scans
  e.g. `scbctl scan amass enum -d example.com` to directly create a `Scan` with a scanType `amass` and the parameters: `["enum", "-d", "example.com"]`. Ideally the cli would also have a paramters to automatically "follow" the scans progress to view the logs of scanner & parser and show a summary of the results.
- [x] directly trigger new executions of `ScheduledScan`s 
  e.g. `scbctl trigger  --namespace internal-scans daily-network-scan` to directly trigger a new Scan for the `daily-network-scan` Scheduled Scan
- [x] interact with cascading rules 

**Additional context**

A simple implementation could be possible using [Cobra](https://github.com/spf13/cobra)

For a previous prototype version of the secureCodeBox there used to be a command line with the ability to start and observe scans, a example scan run can be seen in this video 😛:

https://user-images.githubusercontent.com/13718901/226719918-56d80229-8367-4377-ab1d-9b6761197c69.mp4

J12934:
I dont think the cli should handle the install steps.
These are already pretty straight forward using helm, creating a custom cli for it seems like a big overhead.

The creating / starting scans seems like a more important usecase to me, as the starting of scans can currently be quite cumbersome, as it requires a scan manifest in the localfiles system. Having a "simple cli" with a "scan" command e.g. `scbctl scan nmap scanme.nmap.org` would be pretty cool.

Weltraumschaf:
We have a convenience script for installation though.

Freedisch:
Hello everyone, my name is Thibaut Batale! I'm a final year computer science undergraduate, I'm interested in this project "adding a secureCodeBox CLI" and currently drafting a proposal for it. Just wanted to confirm my thoughts, the main priority commands to build during this term are the create and observe scanners commands.  Followed by update and delete scanners commands ...

J12934:
Hi @Freedisch awesome :)

The prioritise are as described in the issue description.

> - create & observe scans
[...]

PR #2630, opened 2024-08-23
[Feature] Add `--follow` flag to scans command

Freedisch:
## Description

This PR adds a follow flag to the scan command #189 

### Checklist


scbctl/cmd/scans.go
```diff
@@ -116,6 +126,74 @@ func NewScanCommand() *cobra.Command {
...
+
+		if job == nil {
+			fmt.Println("No matching job found, retrying...")
+			time.Sleep(2 * time.Second)
+			continue
+		}
+
+		jobName := job.Name
+		containerName := scanName
+
```
J12934, line 177: The scanName isn't equal to the container name. At least not in all cases.
The "primary" container for which the logs are the most interessting, and which the SCB considers to be the deciding factor if the scan is finsihed is always the first one, so it should be

```suggestion
		containerName := job.Spec.Template.Spec.Containers[0].Name
```

scbctl/cmd/scans.go
```diff
@@ -116,6 +126,74 @@ func NewScanCommand() *cobra.Command {
...
+		fmt.Printf("Found %d job(s)\n", len(jobList.Items))
+
+		for _, j := range jobList.Items {
+			fmt.Printf("Job: %s, Labels: %v\n", j.Name, j.Labels)
+		}
+
+		var job *batchv1.Job
+		for _, j := range jobList.Items {
+			fmt.Printf(j.Name)
+			if strings.HasPrefix(j.Name, fmt.Sprintf("scan-%s", scanName)) {
```
J12934, line 163: The better way to do this is (simmilarly to the cascading command) look for the job by using the owner references of the job to see which actually belong to the scan instead of doing a prefix matching.

scbctl/cmd/scans.go
```diff
@@ -116,6 +126,74 @@ func NewScanCommand() *cobra.Command {
...
 	scanCmd.Flags().String("name", "", "Name of the created scan. If no name is provided, the ScanType will be used as the name")
+	scanCmd.Flags().Bool("follow", false, "Follow the log output")
 
 	return scanCmd
 }
+
+func followScanLogs(ctx context.Context, kubeclient client.Client, namespace, scanName string) error {
+	// Find the job associated with the scan
+	jobList := &batchv1.JobList{}
+	labelSelector := client.MatchingLabels{}
```
J12934, line 137: The scan job all a label on them.


```suggestion
	labelSelector := client.MatchingLabels{
        "securecodebox.io/job-type": "scanner"
    }
```

This way it wouldn't have to go through the whole list of jobs, but only scan jobs.
Which also ties in nicely witht he other suggestion below (owner reference) because this way you know for sure that it is the scanner job of the scan.

scbctl/cmd/scans.go
```diff
@@ -116,6 +126,74 @@ func NewScanCommand() *cobra.Command {
...
+		if err := cmd.Run(); err != nil {
+			return fmt.Errorf("error streaming logs: %s", err)
+		}
+
+		break
+	}
+
+	return nil
+}
+func runKubectlLogs(jobName, containerName, namespace string) error {
```
J12934, line 194: Function is unused right now

scbctl/pkg/client.go
```diff
@@ -31,8 +32,9 @@ func (d *DefaultClientProvider) GetClient(flags *genericclioptions.ConfigFlags)
 
 func init() {
 	utilruntime.Must(v1.AddToScheme(scheme))
+  utilruntime.Must(corev1.AddToScheme(scheme))
```
J12934, line 35: Can you run go fmt over this too? :)

J12934 (changes requested):
Some minor comments.

There are also still a lot of debug logs, which should be removed before merging this :)

scbctl/cmd/scans.go
J12934: The hook status is for the hooks not the scan jobs.
I've changed the code to use the labels then the owner references if you want to have a look what i meant :)

J12934 (approved):

J12934:
recreated this pr by cherry-picking the commits onto a new branch: https://github.com/secureCodeBox/secureCodeBox/pull/2668

will close this PR now

PR #2890, opened 2025-02-06
Update operators to latest kubebuilder versions

J12934:
## Description

Update operator and kubernetes auto-discovery to latest kubebuilder and golang version.

### Checklist


auto-discovery/kubernetes/go.mod
```diff
@@ -4,52 +4,52 @@
 
 module github.com/secureCodeBox/secureCodeBox/auto-discovery/kubernetes
 
-go 1.22.0
+go 1.23.0
```
Weltraumschaf, line 7: Why not using 1.23? Latest is 1.23.6.

lurker/go.mod
```diff
@@ -4,50 +4,52 @@
 
 module github.com/secureCodeBox/secureCodeBox/lurker
 
-go 1.22
+go 1.23.0
```
Weltraumschaf, line 7: Why not using 1.23? Latest is 1.23.6.

Weltraumschaf (commented):
See my comments.

auto-discovery/kubernetes/go.mod
J12934: that's what my local go 1.23.5 had set it to automatically.
that the minimun version required, don't see much point insisting on the latest patch version

Weltraumschaf: 1.23.5has a CVE on Linux. Stumbled upon on a private hack-project. I fixed this by setting the version to 1.23, so that 1.23.6 is automatically used. So I suggest to use only the minor in go.mod.

J12934: apparently just for powerpc cpus, which we dont support.
change it to 1.23.6 anyways, so should be ok now

Issue #1158, opened 2022-05-07
🔧 Maintenance: Move hook integration tests to hook directories


SebieF:
When moving to Makefile builds in the CI, all of the scanner integration-tests were also moved to the specific scanner directories.
[Example ssh-scan tests](https://github.com/secureCodeBox/secureCodeBox/commit/dc45669e58c8db67f3005484d558d65b35edc8ab))

However, the hook integration tests are still located in `secureCodeBox/tests/integration-tests/hooks` for the `finding-post-processing` and `notification` hook. I suggest moving them to a new directory like `secureCodeBox/hooks/finding-post-processing/integration-tests` for consistency with the scanner tests.

Weltraumschaf:
@J12934 says there is may be a reason why it is that way, but he can't remember. Anyway it's a good idea to clean this up.

PR #3088, opened 2025-06-16
Migrate Scanner / Parsers to ESM and @kubernetes/client-node v1.x

J12934:
## Description

Migrates all scanner and hooks to use ecma script modules for code im/exports.
This enables us to upgrade to the new @kubernets/client-node v1.x which only supports ESM (As many new modules do)
To be better able to test this, I've revamped / migrated the the local testing setup from makefiles to task files with (imo) way easier usage w/ automatic kind cluster / testing env setup.

for the hooks, I've reactivated the tests which were still in the global test/integration folder, these haven't been running for 2-3 years. There were also some generic integration tests which were also not in CI, these i've deleted because our normal test coverage for hooks and scanners have gotten so got that there is little point to them.

CI is reporting scanner status checks are missing, i've removed a duplicate space in the job name, that causes the confusion, everything should still be here.

Closes https://github.com/secureCodeBox/secureCodeBox/issues/2851
Closes https://github.com/secureCodeBox/secureCodeBox/issues/1158

### Checklist


J12934:
Closes #2851 
Closes #1158

Reet00 (approved):

PR #3122, opened 2025-07-14
Add subfinder scanner (closes #2852)

### LINKED ISSUE #2852: Evaluate subfinder as potential amass replacement
Amass has been causing a lot of issue since the v4 release.
It seems that the general direction of the tool with its persistent database is not really in line with the tools that are easily supported by the scb. There also have been some bugs / features which have made it very hard to use amass as a reliable subdomain scanner in the SCB anymore.

As a alternative we should evaluate [subfinder](https://github.com/projectdiscovery/subfinder) which seems to tick all the boxes for me that I'd need from a subdomain scanner without the hassle.

joel-sass:
## Description
Added a new scanner for subdomains in order to replace amass in future iterations. (closes #2852)

### Checklist


scanners/subfinder/scanner/Dockerfile
```diff
@@ -0,0 +1,13 @@
+# SPDX-FileCopyrightText: the secureCodeBox authors
```
J12934, line 1: A dockerfile for the scanner shouldn't be required as you're using the subfinder image directly :)
You can just delete the scanner folder entirely in that case

scanners/subfinder/parser/package.json
```diff
@@ -0,0 +1,7 @@
+{
```
J12934, line 1: if the parser doesn't need external dependencies I'd also not use a package.json
Not required and it's just additional bloat

scanners/subfinder/parser/parser.js
```diff
@@ -0,0 +1,45 @@
...
+      "location": item.host,
+      "osi_layer": "NETWORK",
+      "severity": "INFORMATIONAL",
+      "attributes": {
+        "addresses": {
+          "ip": item?.ip || null
+        },
+        "domain": item.input,
+        "hostname": item.host,
+        "ip_address": item?.ip || null,
```
J12934, line 28: we've updated (most, i think) scanners a while back to have an attributes.ip_addresses field with an array of ips. (as a host can resolve to multiple ones)

would be good to have the same here too.

scanners/subfinder/parser/parser.js
```diff
@@ -0,0 +1,45 @@
...
+      "identified_at": null,
+      "description": `Found subdomain ${item.host}`,
+      "category": "Subdomain",
+      "location": item.host,
+      "osi_layer": "NETWORK",
+      "severity": "INFORMATIONAL",
+      "attributes": {
+        "addresses": {
+          "ip": item?.ip || null
+        },
```
J12934, line 25: don't understand why this nested field exists in addition to the ip field below

scanners/subfinder/examples/securecodebox.io/active/findings.yaml.license
```diff
@@ -0,0 +1,3 @@
+SPDX-FileCopyrightText: the secureCodeBox authors
```
J12934, line 1: this license file probably doesn't need to exist?
the file already has a license header

scanners/subfinder/examples/securecodebox.io/active/scan.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+apiVersion: "execution.securecodebox.io/v1"
+kind: Scan
+metadata:
+  name: "subfinder-active-securecodebox.io"
+spec:
+  scanType: "subfinder"
+  parameters:
+    - "-active"
+    - "-oI"
```
J12934, line 13: for more "cryptic" parameters in the example:
would be good to have a short explanation what this parameter does, and ideally use the longer form parameter (i think `-ip` in this case. {(onger might not apply here :D)) to reduce confusion

scanners/subfinder/examples/securecodebox.io/active/scan.yaml
```diff
@@ -0,0 +1,15 @@
+# SPDX-FileCopyrightText: the secureCodeBox authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+apiVersion: "execution.securecodebox.io/v1"
```
J12934, line 5: nice examples 🙌
might be good to add an explanation how the active and passive scans differ to make this clearer to people looking at the examples

scanners/subfinder/Chart.yaml
```diff
@@ -0,0 +1,31 @@
...
+name: subfinder
+description: A Helm chart for the subfinder security Scanner that integrates with the secureCodeBox.
+
+type: application
+# version - gets automatically set to the secureCodeBox release version when the helm charts gets published
+version: v3.1.0-alpha1
+appVersion: "v2.7.0"
+kubeVersion: ">=v1.11.0-0"
+annotations:
+  #versionApi: https://api.github.com/repos/REPONAME/REPO/releases/latest
```
J12934, line 15: would be good to have this set.
This field is used by our scb bot to check daily for new releases and automatically updating the scanner.
check some other scanners, e.g. nuclei for the format

J12934 (changes requested):

scanners/subfinder/parser/parser.js
joel-sass: Since this should act as a replacement for amass in the future I reused the format from the amass scanner for compatibility. Here we have:
```
attributes: {
        addresses: {
          ip: ipObj?.address || null,
          cidr: cidrObj?.cidr || null,
          asn: asnObj?.number || null,
          desc: managedByObj?.name || null,
        },
        ip_addresses: ipObj?.address || null,
```

scanners/subfinder/scanner/Dockerfile
J12934: but you are running the image directly :D
the values.yaml file references the official image.
this might work in tests as it builds up the image locally there but if a user installs the chart later on they get the official image.

Probably the easier way to fix this warning is to specify a user in the helm values under podSecurityContext. 
https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod

See runAsUser and runAsGroup. You can specify ids which don't exist in the image. generally everything over 1000 or better over 10.000 is good :)

J12934 (approved):

PR #3249, opened 2025-08-29
Draft for New Landing Page

Weltraumschaf:


J12934:
some thoughts:

- not sure if this makes sense to have this as a random file in a random directory? is that supposed to be temporary?
- I feel like the differences in the bullet points and the use cases is very slim. e.g. the cascading rules are in both, generally i'd like to make the website as easy to udnerstand and follow us possible thats why i tried to visualize the points which i wanted to highlight and provide links for users to learn more. It feels like all the points could have the same
- we've discussed this before, but the "try as service" button is still seems very confusing namewise to me. ATM this isn't really a service but a small demo with a very limited scope and flexibility.
- nitpick: i think the "what make securecodebox so speical" isn't great as a headline. I don't think people look for special tools, but for useful ones.

Weltraumschaf:
>     * not sure if this makes sense to have this as a random file in a random directory? is that supposed to be temporary?
Just wanted the tooling at hand (version control, discussion in PR). My guess: this will be moved into Figma at some point, and then we can delete this file. OTOH we can introduce a dedicated dir for artwork? When I thin about, maybe this makes more sense, then we have a lace to store sticker files, flyer stuff etc.
 
>     * I feel like the differences in the bullet points and the use cases is very slim. e.g. the cascading rules are in both, generally i'd like to make the website as easy to udnerstand and follow us possible thats why i tried to visualize the points which i wanted to highlight and provide links for users to learn more. It feels like all the points could have the same

IMO a concern of phrasing. I tried to do the bullet points as catchy as possible. First iteration. Only doing it visually is not good for SEO. Also there are different types of men (visually and not so visually).

>     * we've discussed this before, but the "try as service" button is still seems very confusing namewise to me. ATM this isn't really a service but a small demo with a very limited scope and flexibility.

Just a "place holder" that we do not forget this element. I would gather feedback from marketing/ux.

>     * nitpick: i think the "what make securecodebox so speical" isn't great as a headline. I don't think people look for special tools, but for useful ones.

Maybe these points makes it for some people not useful... I would gather feedback from marketing/ux.

J12934:
> IMO a concern of phrasing. I tried to do the bullet points as catchy as possible. First iteration. Only doing it visually is not good for SEO. Also there are different types of men (visually and not so visually).

Thats not what i suggested. The things I've included have both text and visual content / examples.
Also semi related: learning styles are a myth: https://www.youtube.com/watch?v=rhgwIhB58PA

Weltraumschaf:
Moved it to a Miro-Board. The sizes and alignments are quite ruff and not really adjusted. IMO this should do a designer for us.

PR #3271, opened 2025-09-11
Avoid confusion in cascading scans between http on port 443

Reet00:
When running a nmap service scan (-sV) https is marked as ssl/http instead of https, which has resulted into problems with our cascadingscan hook. 
Proposed solution: Set tunnel attribute by default to string "none" which can be matched in the cascadingrules

### Checklist


Reet00:
Also opened an [issue ](https://github.com/nmap/nmap/issues/3188)with nmap to verify that this is intentional and not a bug.
nmap/nmap#3188

J12934 (approved):

PR #3281, opened 2025-09-16
Dependabot/gradle/hooks/persistence defectdojo/hook/gradle version updates 27032e4d85

Weltraumschaf:
For unknown reason the dependency `jackson-annotations` could not be found. Anyway, since Jackson provides a BOM dependency, this is preferable because this ensures consistent versions over all Jackson modules.


Weltraumschaf:
Relates to https://github.com/secureCodeBox/secureCodeBox/pull/3299

J12934 (approved):

PR #3340, opened 2025-10-26
Fix secret name in helm template

yyvfuruta:
## Description

This PR fixes the name of the MinIO secret used in the Manager Deployment.


J12934:
Hi @yyvfuruta ,
sorry for the delay lost this PR out of my sights.
There were duplicate commits in this branch and only one had the signoff on it (not super sure how this is even possible :D)
Should be good to go now :)

PR #3452, opened 2026-01-08
Add option to include target domain in subfinder findings

p4trickweiss:
### Checklist

### DIFF
--- a/scanners/subfinder/parser/parser.js
+++ b/scanners/subfinder/parser/parser.js
@@ -34,7 +34,9 @@ function extractDomainFromArgs(args) {
       if (arg.startsWith(`${flag}=`)) {
         const value = arg.slice(flag.length + 1);
         // Remove surrounding quotes if present
-        return value.replace(/^["']|["']$/g, "");
+        return value
+          .replace(/^["']/, "") // Remove leading quote
+          .replace(/["']$/, ""); // Remove trailing quote
       }
     }
 


scanners/subfinder/parser/parser.js
```diff
@@ -2,40 +2,75 @@
...
+    ? getArgValue(scan.spec.parameters, "-d", "-domain")
+    : null;
+  if (domain) {
+    findings.push(getTargetDomainFinding(domain));
+  }
+
+  return findings;
+}
+
+function getArgValue(args, ...flags) {
```
Reet00: Is there a reason you are retrieving the domain form the arg value? The raw result from the scanner also returns the orginal domain  in each finding. so the domain is also retrievable by `const domain = targets[0].input` .

scanners/subfinder/values.yaml
```diff
@@ -5,6 +5,9 @@
 # -- Define imagePullSecrets when a private registry is used (see: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/)
 imagePullSecrets: []
 
+# includeTargetDomain -- include target domain in findings
+includeTargetDomain: false
```
Reet00: Since we are appending this to env anyways, could'nt we also just do sth like 
env: 
  -name: INCLUDE_TARGET_DOMAIN
    value: false
?

PR #3573, opened 2026-03-25
Fix cascading-hooks labels

aveyrenc:
When supplying multiple labels for the cascading scans hook, we end up with an indention error.

Values:

```yaml
hook:
  labels:
    app: securecodebox
    hooks: cascading
```

Result:

```yaml
kind: ScanCompletionHook
metadata:
  name: release-name-cascading-scans
  labels:
    helm.sh/chart: cascading-scans-5.6.0
    app.kubernetes.io/name: cascading-scans
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/managed-by: Helm
    securecodebox.io/internal: "true"
    app: securecodebox
hooks: cascading
spec:
  priority: 100
  type: ReadOnly
  ```
  
  As you can see, the `hooks: cascading` label is not indented.
  
This PR fixes labels indentation

### DIFF
--- a/hooks/cascading-scans/templates/cascading-scans-hook.yaml
+++ b/hooks/cascading-scans/templates/cascading-scans-hook.yaml
@@ -10,7 +10,7 @@ metadata:
     {{- include "cascading-scans.labels" . | nindent 4 }}
     securecodebox.io/internal: "true"
     {{- with .Values.hook.labels }}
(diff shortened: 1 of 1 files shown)


J12934 (approved):
Hi @aveyrenc sorry lost this out of my sight, looks good 👍

PR #3786, opened 2026-08-25
Add golang based hook-sdk

J12934:
## Description

Also migrate fpp hook to have a test / example golang hook

We might migrate more hooks to golang in the future.

### Checklist


documentation/docs/api/finding.md
```diff
@@ -20,7 +20,7 @@ The 'findings.json' file that contains these Findings complies with the followin
   "$defs": {
     "finding": {
       "type": "object",
-      "additionalProperties": true,
+      "additionalProperties": false,
```
Weltraumschaf, line 23: Was this an old documentation bug?

hook-sdk/golang/sdk.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+		client.k8sClient = k8sClient
+	}
+	if client.fileClient == nil {
+		client.fileClient = NewFileClient(client.logger)
+	}
+	return client, nil
+}
+
+func (c *Client) Run(ctx context.Context, handler HookHandler) error {
```
Weltraumschaf, line 71: Not sure if this is idiomatic Go. Looks like plain conversion from JS (typical pattern in JS to implement "private" implementation inside a closure). I suggest to just introduce a private implementation struct which receives these inner functions and create an Instance with an public constructor function returning the interface struct type.

hooks/finding-post-processing/hook/handler.go
```diff
@@ -0,0 +1,226 @@
...
+	for _, condition := range conditions {
+		if matches(finding, condition) {
+			return true
+		}
+	}
+	return false
+}
+
+func matches(finding hooksdk.Finding, condition findingPatch) bool {
+	if condition.ID != nil && *condition.ID != finding.ID || condition.IdentifiedAt != nil && (finding.IdentifiedAt == nil || *condition.IdentifiedAt != *finding.IdentifiedAt) || condition.ParsedAt != nil && *condition.ParsedAt != finding.ParsedAt || condition.Name != nil && *condition.Name != finding.Name || condition.Description != nil && (finding.Description == nil || *condition.Description != *finding.Description) || condition.Category != nil && *condition.Category != finding.Category || condition.Severity != nil && *condition.Severity != finding.Severity || condition.Mitigation != nil && (finding.Mitigation == nil || *condition.Mitigation != *finding.Mitigation) || condition.Location != nil && (finding.Location == nil || *condition.Location != *finding.Location) || condition.OSILayer != nil && *condition.OSILayer != finding.OSILayer {
```
Weltraumschaf, line 101: Agree with Sonar 😂

hooks/finding-post-processing/hook/handler.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+	if condition.References != nil && !matchesReferences(finding.References, *condition.References) {
+		return false
+	}
+	return true
+}
+
+func matchesScan(scan hooksdk.FindingScan, patch findingScanPatch) bool {
+	return (patch.CreatedAt == nil || *patch.CreatedAt == scan.CreatedAt) && (patch.Name == nil || *patch.Name == scan.Name) && (patch.Namespace == nil || *patch.Namespace == scan.Namespace) && (patch.ScanType == nil || *patch.ScanType == scan.ScanType)
+}
```
Weltraumschaf, line 118: Also quite hard to read. Suggest:
```go
func matchesScan(scan hooksdk.FindingScan, patch findingScanPatch) bool {
	if patch.CreatedAt != nil && *patch.CreatedAt != scan.CreatedAt {
		return false
	}
	if patch.Name != nil && *patch.Name != scan.Name {
		return false
	}
	if patch.Namespace != nil && *patch.Namespace != scan.Namespace {
		return false
	}
	if patch.ScanType != nil && *patch.ScanType != scan.ScanType {
		return false
	}
	return true
}
```

parser-sdk/nodejs/findings-schema.json
```diff
@@ -8,7 +8,7 @@
   "$defs": {
     "finding": {
       "type": "object",
-      "additionalProperties": true,
+      "additionalProperties": false,
```
Weltraumschaf, line 11: See comment on docs.

Weltraumschaf (changes requested):

documentation/docs/api/finding.md
J12934: More of a gap in the existing finding schema, thought we were good to toggle it on, but did a more thourough check and we still have two parsers which set invalid properties, opened up a follow up to fix them in a breaking release: #3797

hook-sdk/golang/sdk.go
J12934: should be resolved

hooks/finding-post-processing/hook/handler.go
```diff
@@ -0,0 +1,265 @@
...
+func matchesAny(finding hooksdk.Finding, conditions []findingPatch) bool {
+	for _, condition := range conditions {
+		if matches(finding, condition) {
+			return true
+		}
+	}
+	return false
+}
+
+func matches(finding hooksdk.Finding, condition findingPatch) bool {
```
Weltraumschaf, line 100: sonar complains because too long. Suggest `//NOSONAR` or something like that to deliberately ignore it.

hook-sdk/golang/sdk.go
Weltraumschaf: What I meant: Why is e.g. `getRawResults()` and all others not "normal" method of `Client` like:
```go
func (Client* c) getRawResults() (string, error) {
  ...
}
```

Weltraumschaf (approved):

